Issue #25, opened 2018-05-08
Histogram component to contain only numeric data


ivanmalagon:
### Context

Our current histogram implementation is in reality a Bar chart. Histograms are a representation of the distribution of numerical data.

To reflect its authentic nature we need to make several changes:

#### Input data

Histogram component need to accept as data an array of buckets with this info:

```
data: [{
  start: number, // The lower limit of the bin
  end: number, // The upper limit of the bin
  value: number // To value represented in the bin
}]
```

The validations to do are to check that everything is a number.

#### X axis

Initial feedback is telling us that people want to see the X axis values. Taking a similar one designed in Airship sketch as an example.

![image](https://user-images.githubusercontent.com/1078228/39756936-89392916-52cb-11e8-9570-92fb3b2edbdf.png)

From that design, I assume that there's no vertical lines in Airship design.

Are you ok with this, @urbanphes ? Drawing x axis values and not showing vertical lines. If not, can you provide us with an example?

#### Documentation

Update documentation

#### Extra ball: bar chart

Since we're getting rid of the histogram component as a bar chart, can we use the current StackedBar as a replacement, @rubenmoya ?

I mean, using it with only one category would behave as the current histogram, right? If so, we can create a component called Bar chart and extend it to Stacked Bar chart.

rubenmoya:
To be honest, I don't know, we should talk about what data formatting we want to use in the `StackedBar` component and maybe we can reuse it. The implementation is quite different from one to another because of the multiple bars.

PR #27, opened 2018-05-09
Change <Histogram /> data format and add bottom axis

rubenmoya:
### Description
Change data format in `<Histogram />` widget.

Before:

```
[
  { name: 'A New Hope', value: 359029623 },
  { name: 'The Empire Strikes Back', value: 236513856 },
  { name: 'Return of the Jedi', value: 204338075 },
]
```

After

```
[
  { start: 8, end: 48, value: 6318 },
  { start: 49, end: 88, value: 5922 },
  { start: 89, end: 128, value: 2363 },
]
```

The `start` and `end` keys are used to calculate the bottom axis, while the `value` is used for the left axis and the tooltip. 

## Screenshots:
screen shot 2018-05-09 at 14 43 51

Related to: #25


docs/components/histogram.md
```diff
@@ -13,13 +13,13 @@ Talk about the Histogram
...
-Value that will be shown in the widget. Each category should include a `name` and a `value`. For example:
+Value that will be shown in the widget. Each bin should include `start`, `end` and `value` keys. For example:
 
 ```
 [
-  { name: 'A New Hope', value: 359029623 },
-  { name: 'The Empire Strikes Back', value: 236513856 },
-  { name: 'Return of the Jedi', value: 204338075 },
+  { start: 8, end: 48, value: 6318 },
+  { start: 49, end: 88, value: 5922 },
```
ivanmalagon: Limits are inclusive. I mean, one bin has `end: 48` and the next one `starts: 48`.

ivanmalagon: Since we're relying in the code on having the input data sorted, let's reflect that in the documentation.

`Bin data should be provided sorted and complete:
  - Sorted: bin order must respect data order. `bin[0]` should start before bin[1] and so on.
  - Complete: if a bin doesn't have any value, it must be provided anyway with the proper `start` and `end` properties and a `value: 0`.

rubenmoya: Done!

PR #35, opened 2018-07-04
Feature stacked bar map

MtlRaze35:
### Description
Interactive Stacked Bar Complete

New Prop added:
selectedData- Function to be used in Prototype demo

### DIFF
--- a/src/components/Popup/popup.js
+++ b/src/components/Popup/popup.js
@@ -26,6 +26,8 @@ const Content = styled.div`
   padding: 16px;
   color: ${props => props.theme.type01};
   font: 400 12px/20px 'Roboto';
+  max-height: 200px;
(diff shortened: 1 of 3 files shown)


ivanmalagon:
We're changing the approach of Airship. After gathering feedback from several partners and users, we are recoding the widgets as Web Components and providing a CSS framework for layout and simple components. In our tests this has been easily integrated with React, Angular or vanilla and will help us to broaden our partners support. A React components-only library has not worked as well as we thought initially.

That means that the current Airship repository is stale and no further development on React is expected. Feel free to fork the current project and make whatever change you need in case you need the widgets as React components.

Interactivity and highlight will be taken into account into the refactor. Thanks for your on it.

Sorry for the inconveniences.

PR #50, opened 2018-07-17
Change some configurations

jesusbotella:
TBD.


ivanmalagon:
Some comments:
- We use semistandard all over the repos, but the eslint conf extends airbnb.
- Do we need to have the same info duplicated in every package.json? At least we should have the same. (Different licenses, yarn / npm, ...)

Issue #53, opened 2018-07-17
Configure CI


IagoLast:
Summary
---

Configure CircleCI to run:
* [x] linter
* [x] tests for components
* [x] test for styles

This depends on #46 #51 #52

PR #55, opened 2018-07-17
Configure CircleCI

VictorVelarde:
Related to #53

### Tasks
Added CircleCI v2 configuration with 3 parallel jobs:
- [x] circleci: **lint**
- [x] circleci: **test-unit:components**
- [x] circleci: **test-unit:styles**

### Current status
Tasks are launched correctly, and all **3 pass** (including test-unit:styles). It will fail in the future, when new reference images will be created (follow manual steps to fix the problem: snapshots generated at Mac local dev and remote-linux CircleCI machine may differ)

---
For some updated notes on CircleCI and reference images creation, see **DEVELOPERS.md**


VictorVelarde:
Currently with just one `reference.png`, manually copied from CircleCI

IagoLast:
Nice job!!!  Could you add a DEVELOPERS.md with a testing section containing the issue comments?

packages/styles/package.json
```diff
@@ -21,7 +21,7 @@
     "lint": "stylelint 'src/**/*.scss'",
     "serve": "serve -p 5000",
     "test:watch": "echo 'No tests available yet.'",
-    "test": "node src/test.js"
+    "test": "npm run build && node src/test.js"
```
IagoLast: Discussion here:

we can force the build here or create a separated step in the circleCI config.

if we create a separated step the test output wont containg build traces (red mark):


screen shot 2018-07-19 at 12 39 21

VictorVelarde: That seems reasonable, I'll create a different step in the job

DEVELOPERS.md
```diff
@@ -0,0 +1,27 @@
+# Developer Notes
```
jesusbotella: ✨

Issue #51, opened 2018-07-17
Basic Component Testing


IagoLast:
Summary
---

We need to find a way to test web components. In this issue, the idea is to tackle unit testing by configuring the system provided by stencil. 

- [x] Implement a basic unit test for the `toggle` component
- [x] Pass the test
- [x] Add a lerna command to run the unit tests `test:unit`

PR #58, opened 2018-07-18
Web Components testing

jesusbotella:
Related to https://github.com/CartoDB/airship/issues/51

This PR sets up component testing with Jest. It was already configured in Stencil, but as we use multiple packages I needed to change the path of a file.

There are two new npm scripts:
- `test:components`: which launches Web Component tests only, and check them once.

- `test:components:watch`: which launches Web Component tests only, but it passes the test suite every time a file is modified. There's a mini GUI with multiple options, but by default, it only runs the tests compelling to the files which its code has been modified since last commmit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,10 @@
   "scripts": {
     "build": "lerna run --parallel build",
     "dev": "lerna run --parallel dev",
-    "lint:fix": "lerna run lint:fix",
+    "test:components": "lerna exec --scope @carto/airship-components -- npm run test",
+    "test:components:watch": "lerna exec -- npm run test:watch",
     "lint": "lerna run lint",
+    "lint:fix": "lerna run lint:fix",
     "postinstall": "lerna bootstrap --hoist",
     "serve": "lerna run --parallel serve",
     "test:watch": "lerna run --parallel test:watch",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -13,11 +13,13 @@
   "scripts": {
     "build": "lerna run --parallel build",
     "dev": "lerna run --parallel dev",
-    "lint:fix": "lerna run lint:fix",
+    "test:components": "lerna exec --scope @carto/airship-components -- npm run test",
```
VictorVelarde: 👍I guess `exec` allows scope, and that's why we don't use `run` here...

jesusbotella: Yes, it does for sure. But if we use run, it doesn't show the colored output, which is so useful. 😞

VictorVelarde:
@jesusbotella LGTM

jesusbotella:
Let's ask @IagoLast and @ivanmalagon for review!

ivanmalagon:
LGTM

I like how similar is the API to Jasmine, so no big learning curve ahead. 👍

jesusbotella:
Yes, Jest is so good :)
The API is similar to Jasmine, but with some improvements. And it runs the test suite in parallel.  🎉

Issue #52, opened 2018-07-17
Basic Style Testing


IagoLast:
Summary
---

Investigate and implement a testing proposal for style/elements

- [x] Implement a dummy element `as-button`
- [x] Create an automated test
- [x] Pass the test

PR #61, opened 2018-07-18
Add exquisite testing

IagoLast:
#52 
---

Use `exquisite-sst` to run screenshot tests.

This takes 3.8s to run 4 test cases 🚀


VictorVelarde:
I can install the dependencies and run it fine, but the test always fails... is that expected?

packages/styles/src/button/button.spec.js
```diff
@@ -0,0 +1,32 @@
...
+const input = path.resolve(__dirname, 'img/reference.png');
+// Path where the screenshot will be saved
+const output = path.resolve(__dirname, 'img/button.png');
+
+// Url to take the screenshot from
+const url = `file://${path.resolve(__dirname, 'button.html')}`;
+
+console.log('Visual regression testing:');
+// Take the screenshot and compare it against the reference image.
+exquisite.test({ input, output, url, delay })
```
VictorVelarde: This test always fails, am I missing anything?

packages/styles/src/button/button.spec.js
```diff
@@ -0,0 +1,32 @@
+const assert = require('assert');
+const path = require('path');
+const fs = require('fs');
+const exquisite = require('exquisite-sst');
+const colors = require('colors');
+
+// Wait 2000 ms after the loaded event to take the screenshot
+const delay = 1000
```
ivanmalagon: can we rely on a event instead of setting a fixed amount of time?

packages/styles/src/button/button.spec.js
```diff
@@ -0,0 +1,32 @@
...
+
+// Reference image
+const input = path.resolve(__dirname, 'img/reference.png');
```
ivanmalagon: how do we take the reference screenshot? With Exquisite a first time?

packages/styles/src/button/button.spec.js
```diff
@@ -0,0 +1,32 @@
...
+const url = `file://${path.resolve(__dirname, 'button.html')}`;
+
+console.log('Visual regression testing:');
+// Take the screenshot and compare it against the reference image.
+exquisite.test({ input, output, url, delay })
+  .then(differentPixels => {
+    // Delete the screenshot
+    fs.unlinkSync(output);
+    // Assert the result
+    assert.equal(differentPixels, 0);
```
ivanmalagon: I assume that we can have a threshold for differences based on what I've heard in CARTO VL, right?

packages/styles/src/button/button.spec.js
IagoLast: yep

packages/styles/src/button/button.spec.js
IagoLast: yep

packages/styles/src/button/button.spec.js
IagoLast: yep

packages/styles/src/button/button.spec.js
IagoLast: missing anything

PR #62, opened 2018-07-18
Setup development server and live reload

jesusbotella:
This PR adds `developer happiness` when it comes to developing Airship, whether you're developing components or styles.

To make it work you just have to execute: `npm run dev` and it will launch a browser pointing to the main index.html entry, located in `packages/components/index.html`.

So, the way it works is:
- It builds and copies Web Components and the styles to `www/` folder.
- Copies all `.html` files within `components/src` and `styles/src` to the proper folders in `www/`.
- Launches the development server pointing to the main `index.html`.

Whenever you change any file within the components source code, the browser will reload the page. But when changing styles, the browser won't reload (😞), so we'll have to press `CMD + R` to refresh the page.

Feedback is welcome! :)


jesusbotella:
I have updated the PR if you want to take a look at it.

I made a change to copy the component examples with Stencil dev server, and the styles examples are copied with a custom script because I couldn't make it work with Stencil development server :(

IagoLast:
Found a weird thing:

It looks the livereload could lead to uncontroled state in components:

1. npm i
2. npm run dev
3. Open http://localhost:3333/components/as-switch/as-switch.html
4. Click in the switch so the ball goes to the right
5. Make some changes in the css
6: Weird stuff:

screen shot 2018-07-19 at 12 28 59

IagoLast:
When developing `styles` the stylesheets are not found!

screen shot 2018-07-19 at 12 31 28

packages/components/styles/src/button/button.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
```
IagoLast: I dont understand... why is the styles folder duplicated under the components folder?

Issue #64, opened 2018-07-19
Release process


IagoLast:
We are now in an alpha stage, but we need to define how airship is going to be released.

- [x] Discuss about versioning
- [x] Release a `0.0.0-apha-26` just for testing

PR #66, opened 2018-07-19
64 release process

IagoLast:
#64 
---

This adds a new script:

    npm run release

this script has all we need to publish new airship versions


packages/components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@carto/airship-components",
-  "version": "0.0.0",
+  "version": "1.0.0-alpha.26",
```
VictorVelarde: Reading lerna docs, it seems that we could skip individual version for the packages, as we would follow **fixed mode** (both packages always coordinated): https://www.npmjs.com/package/lerna#fixedlocked-mode-default Right?

IagoLast: I think fixed mode makes all packages to have the same version

VictorVelarde: Correct, didn't we want just that @jesusbotella ?

IagoLast: yes :)

jesusbotella: Yep! 😉

jesusbotella:
I am kinda spoiled by prettifier or whatever it is, because I have seen that we changed the properties' order in `package.json` in several PRs. 😞

IagoLast:
so can we 🚀  this?

IagoLast:
@VictorVelarde what do you mean? the version is auto-inyected by lerna

Issue #68, opened 2018-07-19
TsLinting errors wont make Lerna fail


IagoLast:
When you run `npm run lint` on the top level and tslint detects some error the lerna returns a `success` status.

This should fail instead

```
> lerna run lint

lerna info version 2.11.0

> @carto/airship-style@0.0.0 lint /Users/iago/Workspace/airship/packages/styles
> stylelint 'src/**/*.scss'


> @carto/airship-components@0.0.0 lint /Users/iago/Workspace/airship/packages/components
> npm run lint:ts & npm run lint:css


> @carto/airship-components@0.0.0 lint:ts /Users/iago/Workspace/airship/packages/components
> tslint -c tslint.json src/**/*.tsx src/**/*.ts


> @carto/airship-components@0.0.0 lint:css /Users/iago/Workspace/airship/packages/components
> stylelint 'src/components/**/*.scss'


ERROR: src/components/as-switch/as-switch.tsx[5, 3]: The key 'styleUrl' is not sorted alphabetically
ERROR: src/components/as-switch/as-switch.tsx[5, 12]: missing whitespace
ERROR: src/components/as-switch/as-switch.tsx[14, 21]: " should be '
ERROR: src/components/as-switch/as-switch.tsx[15, 21]: " should be '

lerna success run Ran npm script 'lint' in packages:
lerna success - @carto/airship-components
lerna success - @carto/airship-style
```

Issue #67, opened 2018-07-19
Tslint not working


IagoLast:
Our current linting task only lints .ts files

  "lint:ts": "tslint -c tslint.json 'src/**/*.ts'",

but we got `.tsx` files so is not working

PR #70, opened 2018-07-19
Fix Tslint issues

IagoLast:
#67 #68
---

  - Make lerna fail when some script fails
  - Fix some linting issues in files


packages/components/package.json
```diff
@@ -12,10 +12,10 @@
   "scripts": {
     "build": "stencil build",
     "dev": "stencil build --dev --watch --serve",
-    "lint:css-fix": "stylelint 'src/components/**/*.scss' --fix",
+    "lint:css:fix": "stylelint 'src/components/**/*.scss' --fix",
     "lint:css": "stylelint 'src/components/**/*.scss'",
-    "lint:fix": "npm run lint:css-fix",
-    "lint:ts": "tslint -c tslint.json 'src/**/*.ts'",
+    "lint:fix": "npm run lint:css:fix",
+    "lint:ts": "tslint -c tslint.json src/components/**/*.ts src/components/**/*.tsx",
```
jesusbotella: We can use `src/components/**/*.{ts,tsx}` instead of both relative paths

package.json
```diff
@@ -16,8 +16,8 @@
     "dev": "lerna run --parallel dev",
     "test:components": "lerna exec --scope @carto/airship-components -- npm run test",
     "test:components:watch": "lerna exec --scope @carto/airship-components -- npm run test:watch",
-    "lint": "lerna run lint",
-    "lint:fix": "lerna run lint:fix",
+    "lint": "lerna exec npm run lint",
```
jesusbotella: lerna forwards error code by default. I suspect that the failure comes from https://github.com/CartoDB/airship/blob/master/packages/components/package.json#L19. Shouldn't it be `&&`?

packages/components/src/components/as-switch/as-switch.tsx
```diff
@@ -1,23 +1,24 @@
 import { Component, Event, EventEmitter } from '@stencil/core';
 
 @Component({
+  shadow: false,
```
jesusbotella: Are all these changes suggested by the linter? 😂

packages/styles/package.json
```diff
@@ -21,7 +21,8 @@
     "lint": "stylelint 'src/**/*.scss'",
     "serve": "serve -p 5000",
     "test:watch": "echo 'No tests available yet.'",
-    "test": "node src/test.js"
+    "test": "node src/test.js",
```
jesusbotella: I'd do this the other way around.

```
"test": "npm run test:styles",
"test:styles": "node src/test.js"
```

What do you think?

package.json
IagoLast: 🤦‍♂️ totally right

packages/components/src/components/as-switch/as-switch.tsx
IagoLast: yep, vscode linter

packages/styles/package.json
IagoLast: +1

package.json
VictorVelarde: With **&** it should launch both processes in parallel. If **&&** gets along with lerna, then let's use it and go for the serial execution

jesusbotella: Yes, but I think `&` doesn't forward the error code properly to npm 🤷‍♂️

packages/styles/package.json
VictorVelarde: That's fine for me. I started with that indeed, and then simplified it. Are we expecting more type of tests here?

package.json
```diff
@@ -16,8 +16,8 @@
     "dev": "lerna run --parallel dev",
     "test:components": "lerna exec --scope @carto/airship-components -- npm run test",
     "test:components:watch": "lerna exec --scope @carto/airship-components -- npm run test:watch",
-    "lint": "lerna exec npm run lint",
-    "lint:fix": "lerna exec npm run lint:fix",
+    "lint": "lerna run lint",
```
VictorVelarde: Do we lose colored console with this?

IagoLast: no idea

package.json
IagoLast: yes, we ned to use &&

packages/styles/package.json
```diff
@@ -20,9 +20,9 @@
     "lint:fix": "stylelint 'src/**/*.scss' --fix",
     "lint": "stylelint 'src/**/*.scss'",
     "serve": "serve -p 5000",
-    "test:watch": "echo 'No tests available yet.'",
-    "test": "node src/test.js",
-    "test:styles": "npm run test"
```
jesusbotella: Please please please, be careful with prettifier, everything is broken now because of the non-allowed dangling comma in this package.json.

packages/components/package.json
```diff
@@ -12,10 +12,10 @@
   "scripts": {
     "build": "stencil build",
     "dev": "stencil build --dev --watch --serve",
-    "lint:css-fix": "stylelint 'src/components/**/*.scss' --fix",
+    "lint:css:fix": "stylelint 'src/components/**/*.scss' --fix",
```
VictorVelarde: Much better!

package.json
VictorVelarde: so be it

VictorVelarde:
I removed the offending comma and linter is successful

PR #81, opened 2018-07-20
Component Showcase

### LINKED ISSUE #57: Component Showcase
Summary
---

We need a place to showcase all Airship components, including all the possible variants of the components. It can serve as a development tool, because it will include the components bundle as well as the styles.

jesusbotella:
This PR adds [Catalog](https://www.catalog.style/) as our components and styles showcase. We were using it with React as well, but I took another approach as I didn't want to add React to our dependencies. We use Catalog from a JS script located in their CDN.

The approach I followed is to include the Markdown reference from the source code folder of our components/styles, so that we can use it in GitHub as well. The only downside is that we need to call them `README.md` because of GitHub.

Closes #57.


catalog/intro.md
```diff
@@ -0,0 +1,13 @@
+Minimal UI Kit for CARTO solutions, with reusable components.
+
+## Installation
+Airship is available as an [npm package](https://www.npmjs.com/package/@carto/airship).
```
VictorVelarde: How about saying "two npm packages: `@carto/airship-styles` and `@carto/airship-components`?

Issue #63, opened 2018-07-19
Setup a script to generate and download image references from CircleCI


IagoLast:
The visual regression references should be generated in the CI environment.

Currently, the way to go is to upload the code, run the CI with ssh and manually download the references using [scp](https://en.wikipedia.org/wiki/Secure_copy).

We need to improve this process. Some suggestions:

- Create a script that can be run on the CI which generates the references and creates a new PR/commit with them.
- Create a script to easy scp the references from the CI and manually add the new references to the repo

PR #82, opened 2018-07-20
Generate references from CI

IagoLast:
Fixes: #63
---

CircleCI currently passes all checks. But it will fail when new *reference images* will be created for our tests.

Follow these manual steps to fix the problem, as snapshots generated at local dev and remote CircleCI machine may differ, and we want CircleCI ones to be the authoritative source.

#### With this PR those images would be automatically generated by our CI:

1. Create your component.
2. If you want circleCI to generate the reference generate a`ìmg` directory with a empty `.gitkeep` file inside.
3. Open a PR/commit
4. CircleCI will generate a branch with the same name adding the new reference in a commit.
5. Merge this branch into yours when ready.


jesusbotella:
I changed the base branch to the one in the script so that we can see the changes made only in this branch

Issue #38, opened 2018-07-10
Style framework guidelines


ivanmalagon:
For building the Airship CSS framework, these are the guidelines to follow:

- Sass preprocessor.
- BEM methodology.
- Namespacing: for utilities, state classes and JS hooks.
- ITCSS architecture.
- Stylelint.

These are common choices in CARTO.

- Sass preprocessor: used in Builder and Hangar.
- BEM methodology: we've tried to follow it but we are not consistent. Let's do it better this time. You can see it being used in [Hangar](https://github.com/CartoDB/hangar-alpha/blob/master/src/scss/components/_navbar.scss)
- Namespacing for utilities: used in Builder and Hangar. For instance, for positioning we use `.u-tspace-8` and `is-small`, `is-caption` for state. 
- ITCSS: suggested by @elenatorro Let's give it a try.
- Stylelint: used in Builder CI.

### Links

**BEM**
- [BEM introduction](http://getbem.com/introduction/)
- [BEM by example](https://seesparkbox.com/foundry/bem_by_example)
- [BEM 10 common problems](https://www.smashingmagazine.com/2016/06/battling-bem-extended-edition-common-problems-and-how-to-avoid-them/)

**Namespacing**
- [Writing modular CSS - namespaces](https://zellwk.com/blog/css-architecture-2/) Utilities and state applies.

**ITCSS**
- [Building a maintainable and scalable CSS with ITCSS](https://medium.okgrow.com/building-a-maintainable-and-scalable-css-codebase-with-itcss-ceda5b2f495b)
- [CSS at Trivago](https://medium.com/@pistenprinz/css-at-trivago-part-1-structure-and-itcss-52f63ed557ca)
- [CSS at Trivago 2 ](https://medium.com/@pistenprinz/css-at-trivago-part-2-naming-conventions-and-methodologies-d51b445a3a39)

**Stylelint**
- We're using `stylelint-config-standard` and `stylelint-config-property-sort-order-smacss`. Study if there's more config that can helps us with BEM or ITCSS.

Any thoughts @piensaenpixel ?

piensaenpixel:
<3

ivanmalagon:
Two articles to be inspired by:

https://zellwk.com/blog/css-architecture-3/

https://speakerdeck.com/dafed/managing-css-projects-with-itcss?slide=54

PR #97, opened 2018-07-26
Upgrade toolbar styles

IagoLast:
<issue-code>
---


packages/styles/src/core/layout/toolbar/_actions.scss
```diff
@@ -2,14 +2,15 @@
   display: flex;
   position: absolute;
   z-index: 3;
-  top: 52px;
+  top: 13 * $u1;
```
jesusbotella: I wouldn't do this at all. This is hard to read.

Let's use my $header-height variable, and we can multiply there if we want.

packages/styles/src/core/layout/toolbar/_main.scss
```diff
@@ -1,36 +1,39 @@
+$_toolbar-main__size: 13 * $u1;
```
jesusbotella: SASS variables don't need to follow BEM 😂, and they don't need `_` modifier.

And let's put variables in its own folder because if we put this here we don't allow anyone to override it.

packages/styles/src/core/layout/toolbar/_main.scss
```diff
@@ -1,36 +1,39 @@
...
-
-  &--bg-brand-1 {
-    background-color: var(--brand-01-color, $brand-01-color);
-  }
-
-  &--bg-brand-2 {
-    background-color: var(--brand-02-color, $brand-02-color);
-  }
+  height: $_toolbar-main__size;
+  background-color: var(--toolbar-main-bg-color, $brand-02-color);
```
jesusbotella: This is sooo good, we need to define good CSS Variable names to let the users override it.

packages/styles/src/core/layout/toolbar/_main.scss
```diff
@@ -1,36 +1,39 @@
...
 
-  &--bg-brand-3 {
-    background-color: var(--brand-03-color, $brand-03-color);
+  &__toggle {
```
jesusbotella: 🤔 why are you modifying these properties?

packages/styles/src/core/layout/toolbar/_main.scss
```diff
@@ -1,36 +1,39 @@
...
+    position: absolute;
+    z-index: 1;
   }
 
   &__item {
     display: flex;
     align-items: center;
+    justify-content: center;
```
jesusbotella: Let's keep it as I did in my PR, we don't need to force users to use a fixed width/height

packages/styles/src/core/layout/toolbar/_main.scss
```diff
@@ -1,36 +1,39 @@
...
-    height: 48px;
-    min-height: 48px;
-    padding: 8px;
+    padding: 3 * $u1;
+    text-align: center;
     cursor: pointer;
 
     p {
-      margin-left: 8px;
+      margin-left: $u3;
```
jesusbotella: We need to avoid setting CSS properties to bare elements, we need to use classes for this. I changed it in my PR.

packages/styles/src/core/variables/_variables.scss
```diff
@@ -1,2 +1,8 @@
 @import '_colors';
 @import '_breakpoints';
+
+$u1: 4px;
```
jesusbotella: 😞 I don't like this at all. This is hard to understand.

packages/styles/src/core/layout/toolbar/_main.scss
IagoLast: Im not sure what will happen with the components if we put the vars in its own folder 🙄

packages/styles/src/core/layout/toolbar/_main.scss
ivanmalagon: +1 to jbot's comment

packages/styles/src/core/variables/_variables.scss
ivanmalagon: What you don't like? The naming?

packages/styles/src/core/layout/toolbar/_tabs.scss
```diff
@@ -1,37 +1,27 @@
+$_toolbar-height: 13 * $u1;
```
ivanmalagon: Two things:
- If we need `13 * $u1`, let's make `$13u1`.
- Are we using `$toolbar-height` also in widths? When the toolbar is vertical.

jesusbotella: I don't see that `$13u1`, honestly, because we would need to create many variables that might be used once.

And personally, I understand better what `13 * $u1` is, rather than using `$13u1`.

What I don't really like is the `$u1`, but that's a naming thing.

packages/styles/src/core/variables/_variables.scss
jesusbotella: Yes, the naming. I'd say `$base-size`, or `$base-grid-size`, is better.
Anything else that explain better what that variable is.

jesusbotella:
If we want to achieve better accessibility, we might need to change [all these spans](https://github.com/CartoDB/airship/blob/640e8d2801e3b8b9a8e9c4ecf82ba664db629ea1/examples/layouts/generic.html#L36-L51) and [these ones](https://github.com/CartoDB/airship/blob/640e8d2801e3b8b9a8e9c4ecf82ba664db629ea1/examples/layouts/generic.html#L56-L59) to `<li>`

IagoLast:
> If we want to achieve better accessibility, we might need to change all these spans and these ones to <li>

why?

IagoLast:
Thr problem with that is the markup and the css gets more complicated... we need to discuss that

jesusbotella:
This looks totally AWESOME with the latest changes that @IagoLast did ✨

examples/layouts/toolbar.html
```diff
@@ -5,25 +5,91 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Main Layout CDN</title>
+  <title>Generic Layout</title>
   <link rel="stylesheet" href="../../packages/styles/dist/airship.css">
+  <script src="../../packages/components/dist/airship.js"></script>
+
+  <!-- INCLUDE TYPOGRAFY -->
+  <link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet">
```
ivanmalagon: Should we import Roboto via @import? à la `@import url('https://fonts.googleapis.com/css?family=Roboto+Mono');`

Just suggesting. I don't know the best way to do it but I'd appreciate to free the user to include the tag to use Airship as it's thought to look.

packages/styles/src/core/layout/toolbar/_toolbar.scss
```diff
@@ -1,3 +1,41 @@
-@import '_main';
 @import '_actions';
 @import '_tabs';
+
+.as-toolbar {
+  --toolbar-bg-color: $brand-02-color;
```
ivanmalagon: ❤️

packages/styles/src/core/variables/_variables.scss
```diff
@@ -1,3 +1,9 @@
 @import '_colors';
 @import '_breakpoints';
 @import '_sizes';
+
+$u1: 4px;
+$u2: 2 * $u1; // 8
+$u3: 2 * $u2; // 16
+$u4: 2 * $u3; // 32
+$u5: 2 * $u4; // 64
```
ivanmalagon: I don't get making the `$u` exponential.

The grid is 4px based. Where are 12px, 20px, 40px... ?

packages/styles/src/core/layout/README.md
```diff
@@ -4,17 +4,16 @@ Airship is designed to develop Location Intelligence apps without effort. To do
...
+- **Sidebar**: One or two sections beside the map wrapper where the main application contents should be placed.
+- **Map Wrapper**: The area where the map will be placed.
 
 ### Basic Layout
 
+The following example contains an very basic layout with an empty toolbar on top and an `app-container` with a sample CARTO.js map.
+
 ```html
 noSource: true
 responsive: true
```
ivanmalagon: L30: `as-toolbar-main` is outdated.

ivanmalagon: L33: `as-app-container` is outdated.

packages/styles/src/core/layout/README.md
```diff
@@ -40,24 +39,30 @@ collapsed: true
...
+      <img src="https://material.io/tools/icons/static/icons/baseline-fingerprint-24px.svg" alt="Action 1">
+      <p class="as-toolbar__icon-text">Action 1</p>
+    </li>
+    <li class="as-toolbar__item">
+      <img src="https://material.io/tools/icons/static/icons/baseline-add_location-24px.svg" alt="Action 2">
+      <p class="as-toolbar__icon-text">Action 2</p>
+    </li>
+  </ul>
 </nav>
 ```
```
ivanmalagon: I'm missing an explanation of what `openSideMenu()` does. You tell that you need a button but you also need a event handler that does something, right? If you only provide a button, nothing will happen.

ivanmalagon: `as-toolbar-main` is outdate throughout all code examples.

packages/styles/src/core/variables/_variables.scss
IagoLast: otherwise we would need 64 variables...

ivanmalagon: I'm pretty sure that the low ones (4, 8, 12, 16, 20, 24) will be way more frequent that 32 or 64. This exponential growth doesn't reflect the grid steps.

And when doing tiny adjustments you'll go from one value to the nearest one.

Go for
4 = u1 - 1 space
8 = u2 - 2 spaces
12 = u3 - 3 spaces
16 = u4 - 4 spaces

instead of
4 = u1 - 1 space
8 = u2 - 2 spaces
16 = u3 - 4 spaces
32 = u4 - 8 spaces
64 = u5 - 16 spaces

There's no visual relation between the u suffix and the real grid steps.

jesusbotella: I'd go for one base unit variable, such as `$su` (I talked with iago about this), and then multiply wherever we need. 

So that `$toolbar-size: 52px;` would be `$toolbar-size: 13 * $su;`

examples/layouts/toolbar.html
jesusbotella: Yes, I am not sure about it either.

I talked it the other day with @IagoLast but we didn't reach any conclusion. I'd say that I would include it in CSS and that's it, but 🤷‍♂️ . We can talk about it :)

packages/styles/src/core/layout/panels/test/panels.spec.js
```diff
@@ -0,0 +1,20 @@
+const path = require('path');
+
+// Mobile Viewport Settings
+const viewportWidth = 375;
```
IagoLast: 👀

packages/styles/src/core/layout/toolbar/README.md
```diff
@@ -1,4 +1,5 @@
 ### Toolbar
+hola
```
IagoLast: 👀

packages/styles/src/core/layout/toolbar/_actions.scss
```diff
@@ -2,14 +2,22 @@
...
+  top: $toolbar-size;
   right: 0;
   bottom: 0;
   left: -100%;
   flex-direction: column;
+  align-items: flex-start;
   width: 100%;
   transition: left 0.3s;
-  background-color: var(--brand-02-color, $brand-02-color);
+  background-color: var(--toolbar-tabs-bg-color, $brand-02-color);
```
rubenmoya: This won't work on `IE11`, we need a fallback:

```
background-color: $brand-02-color;
background-color: var(--toolbar-tabs-bg-color, $brand-02-color);
```

examples/layouts/generic.html
```diff
@@ -27,80 +27,95 @@
...
   <header class="as-toolbar">
-    <img onclick="_toggleDrawer()" class="as-toolbar__item as-toolbar__menu-toggle" src="https://material.io/tools/icons/static/icons/baseline-menu-24px.svg" alt="">
-    <!-- <span onclick="_toggleDrawer()" class="as-toolbar__item as-toolbar__menu-toggle">
-      <i class="as-icon__hamburger-menu"></i>
-    </span> -->
+    <button onclick="_toggleDrawer()" class="as-toolbar__item as-toolbar__toggle">
+        <img src="https://material.io/tools/icons/static/icons/baseline-menu-24px.svg" alt="">
+    </button>
+    <li class="as-toolbar__item">
+      <svg width="130px" height="36px" viewBox="0 0 92 36" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" style="filter: invert(1);">
```
IagoLast: is it necessary to use inline svg?

packages/styles/src/core/layout/map-wrapper/README.md
```diff
@@ -0,0 +1,71 @@
...
+  </main>
+</body>
+```
+
+
+### Map wrapper + Bottom panel
+
+To add a bottom panel you need to add an element with class `as-panels` as a child of `as-map-wrapper` and an element with class `as-panels__bottom` inside it.
+
+The bottom panel is hidden in mobile and visible in desktop size. The reason is to have the most visible map area possible in small screens. To make the panel visible you need to add the class `is-visible` to the `as-panels` element that contains the `as-panels__bottom` panel.
```
IagoLast: Im not ok with the `is-visible` class.

We agreed to use BEM and this is precisely the given example of when to use a BEM Modifier. I don't understand the reasons for this change.

packages/styles/src/core/layout/map-wrapper/README.md
```diff
@@ -0,0 +1,71 @@
...
+        <div class="as-panels__bottom">
+          Bottom panel content.
+        </div>
+      </div>
+    </div>
+  </main>
+</body>
+```
+
+### Map wrapper + Floating panels
```
IagoLast: missing doc here?

packages/styles/src/core/layout/sidebar/README.md
```diff
@@ -0,0 +1,97 @@
...
+- Position:
+  - `as-sidebar--left`: Puts the sidebar on the left of the map.
+  - `as-sidebar--right`: Puts the sidebar on the right of the map.
+- Size:
+  - `as-sidebar--l`: Makes the sidebar width to be `360px`
+  - `as-sidebar--xl`: Makes the sidebar width to be `460px`
+
+
+
+Sidebars are hidden by default in mobile devices and only will be shown when the modifier class `as-sidebar--left--visible` or `as-sidebar--right--visible`  is present.
```
IagoLast: this is not consistent with the `is-visible`

packages/styles/src/core/layout/map-wrapper/README.md
ivanmalagon: In the original issue for CSS guidelines, the state classes are mentioned:
https://github.com/CartoDB/airship/issues/38

We thought that this is a very common state for everything so it seemed a good idea to have the same syntax for everything (to ease development). Indeed, all the examples turned to be way easier and using less code.

You're right in the specificity issue.

packages/styles/src/core/layout/map-wrapper/README.md
ivanmalagon: Yes. Missing doc. I'll get it.

packages/styles/src/core/layout/toolbar/_actions.scss
ivanmalagon: Yes, we need a fallback. Maybe a PostCSS plugin?

IagoLast: yep, we can wait for the first release to do this kind of fine-tunning

ivanmalagon:
are you all OK with the latest changes?

packages/styles/src/core/layout/map-wrapper/README.md
```diff
@@ -0,0 +1,187 @@
...
+  </main>
+</body>
+```
+
+
+### Map wrapper + Bottom panel
+
+To add a bottom panel you need to add an element with class `as-bottom-bar` as a child of `as-map-wrapper`.
+
+The bottom panel is hidden in mobile and visible in desktop size. The reason is to have the most visible map area possible in small screens. To make the panel visible you need to add the class `as-bottom-bar--visible` to the element that contains the `as-panels__bottom` panel.
```
IagoLast: To make the panel visible `on small screen sizes` ...

IagoLast: > you need to add the class as-bottom-bar--visible to the element that contains the as-panels__bottom panel.

This is not very clear to me

IagoLast:
Two small changes regarding the doc and LGTM

packages/styles/src/core/layout/map-wrapper/README.md
ivanmalagon: Didn't see it on time. I'll do it in my next branch. Sorry.

PR #98, opened 2018-07-26
Refactor toolbar styles

jesusbotella:
Refactor Toolbar styles to use a simpler HTML markup

### DIFF
--- a/packages/styles/src/core/variables/_sizes.scss
+++ b/packages/styles/src/core/variables/_sizes.scss
@@ -1,0 +1,1 @@
+$header-height: 52px;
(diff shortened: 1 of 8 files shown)


jesusbotella:
Fixed ✨

packages/styles/src/core/variables/_sizes.scss
```diff
@@ -0,0 +1 @@
+$header-height: 52px;
```
IagoLast: I think this kind of variables should have a name equal to its component/element, in this case `$toolbar-height`

Issue #75, opened 2018-07-19
Colors


ivanmalagon:
Airship comes with the following color palette:

![image](https://user-images.githubusercontent.com/1078228/42943533-50d47ea6-8b63-11e8-9b31-4221b296dd93.png)

Brand colors
01: #1785FB
02: #0F2D53
03: #47DB99

Type colors
01: #2C2C2C
02: #747474
03: #BABABA
04: #FFFFFF

UI colors
01: #FFFFFF
02: #F5F5F5
03: #E2E6E3
04: #D1D5D7

Support colors
01: #F3522B
02: #FDB32B
03: #80B622

PR #105, opened 2018-07-31
Add basic colors

rubenmoya:
#75 Add basic colors
---

This PR adds the basic Airship color scheme with SASS and CSS variables and updates them in the documentation and elements.

It also adds `postcss-css-variables` which adds a fallback with static content to the rules using CSS variables.

I ended up using `Gulp` since it was a pain to handle the integration between `node-sass` and `postcss`


rubenmoya:
We are defining the variables always in `packages/styles/src/core/core.scss`

IagoLast:
I meant using the [fallback values] in `var`:

```css
background-color: var(--panels-bg-color, $color-primary);
```

rubenmoya:
Yes, but as I said, we are declaring `--panels-bg-color` ourselves, so there is no way it's not defined

rubenmoya:
Nevermind, we need to add it because of the PostCSS plugin, since the variables are defined in core, and it's not available in the standalone components

IagoLast:
> Yes, but as I said, we are declaring --panels-bg-color ourselves, so there is no way it's not defined

an user could f^*ck the styles overriding the variables with javascript :trollface: 

> Nevermind, we need to add it because of the PostCSS plugin, since the variables are defined in core, and it's not available in the standalone

I knew we did it for something!
:trollface:

Issue #90, opened 2018-07-24
Category widget


ivanmalagon:
### Mockup

![image](https://user-images.githubusercontent.com/1078228/43138443-29edee86-8f4f-11e8-8408-469a4631931f.png)

### Props

**Title**
The text to show in the title

**Description**
The description text of the widget

**Show header**
If false, Title and description are not rendered

**Categories**
The categories to show in the widget. The expected object is:
```
{
  name: String. The name of the category,
  value: Number. The number to show along with the category.
  color: String. The color for rendering the category bar. If not present, the default bar color is used.
}
```

**Selected categories**
The categories that are selected.
Format: strings array.

Note: this should be exposed also as a getter. From the outside we need to have access to the selected categories.

**Use total percentage**
If true, we use the total sum of all categories (including others) to render the % in each bar.
If no, we use the max category value to render the %.
Default: `false`

**Default bar color**
This color will be used to draw the bars. It has a default value.

**Show clear button**
Boolean.
If present, it'll show a button when there are any category selected.
The button, when pressed, will clear the selected categories.

![image](https://user-images.githubusercontent.com/1078228/43151464-41eed94c-8f6c-11e8-9688-174029cf5854.png)

### Functions

**clearSelection**
When called, the selected range gets reset. Useful for delegating the filtering logic to another component.

### Events

**categoriesSelected**
Triggered when the selected categories change.
Returns the selected categories array.

### Filtering
This API proposal allows two different filtering scenarios:
1. We want to filter as soon as the user clicks on a category.
    In this scenario, we can listen to `onCategoriesSelected` event, set `showClearButton` to `true` and change the source of the dataview instantly (or wait some time).
    As soon as the filtering occurs, the widget will change its categories, keeping the selected ones active.
    The clear button allows us to have an independent filter on this widget and not worry about other widgets.

2. We want to delegate filtering to some other component. For instance, we want an `Apply all filters` button.
    In this scenario, we can listen to `onCategoriesSelected`. If triggered, we notify to the filter component that there's been a change so we can activate an `Apply all filters` button.
   When the user clicks on it, we go to the Category widget, read the selected categories and then do the source changes.
    The `clear` button is this scenario is optional. 

### Remarks
- `Other` category is not selectable.

PR #107, opened 2018-07-31
Category Widget

jesusbotella:
This PR contains Category Widget as a Web Component. The functionality is described in #90.
Styles and colors were taken from Sketch design.

### Example
Category Widget

### DIFF
--- a/packages/components/src/components/as-category-widget/README.md
+++ b/packages/components/src/components/as-category-widget/README.md
@@ -117,7 +117,7 @@ categoryWidget.useTotalPercentage = true;
 There are some CSS Variables that you can override to change visual styles.
 
 #### **--category-widget--description--color**
-Default: `#1785FB`
+Default: $color-type-20 (`#1785FB`)
 
 ```code
 lang: javascript
@@ -135,7 +135,7 @@ document.body.style.setProperty('--category-widget--bar--height', '8px')
 ```
 
 #### **--category-widget--background-color**
-Default: `#FFF`
+Default: $color-ui-10 (`#FFF`)
 
 ```code
 lang: javascript
@@ -144,7 +144,7 @@ document.body.style.setProperty('--category-widget--background-color', '#F5F5F5'
 ```
 
 #### **--category-bar--background-color**
-Default: `#F5F5F5`
+Default: $color-ui-20 (`#F5F5F5`)
 
 ```code
 lang: javascript
--- a/packages/components/src/components/as-category-widget/README.md
+++ b/packages/components/src/components/as-category-widget/README.md
@@ -3,20 +3,25 @@ Category widget displays the categories passed to the component as a list with a
 ```html
 noSource: true
 ---
-<iframe src="/examples/components/as-category-widget/simple.html" style="width: 100%; height: 330px;">
+<iframe src="/examples/components/as-category-widget/simple.html" style="width: 100%; height: 354px;">
 ```
 
 ```code
 lang: html
 showSource: false
 ---
+<!-- Example with all default values and showing clear button -->
 <as-category-widget
   heading="Business Volume"
-  description="Description"></as-category-widget>
+  description="Description"
+  default-bar-color="#47DB99"></as-category-widget>
 
 <script>
   const categoryWidget = document.querySelector('as-category-widget');
+  categoryWidget.showHeader = true;
   categoryWidget.showClearButton = true;
+  categoryWidget.useTotalPercentage = false;
+  categoryWidget.visibleCategories = Infinity;
   categoryWidget.categories = [
     { name: 'Bars & Restaurants', value: 1000, color: '#FABADA' },
     { name: 'Fashion', value: 900 },
@@ -31,8 +36,10 @@ showSource: false
 
 ### Props
 
-#### **categories** (array required)
-Array of categories, each category should include a `name` and a `value`. You can also override the bar color for each category with `color`. For example:
+#### **categories**: Category[] = []
+Array of categories, each category should include a `name` and a `value`. You can also override the bar color for each category with `color`.
+
+For example:
 
 ```code
 lang: javascript
@@ -43,7 +50,7 @@ categoryWidget.categories = [
   { name: 'Return of the Jedi', value: 204338075 },
 ];
 ```
-#### **defaultBarColor** (string)
+#### **defaultBarColor**: string = '#47DB99'
 Overrides default color to draw the bars. Default value is `#47DB99`.
 
 ```code
@@ -57,7 +64,7 @@ lang: javascript
 categoryWidget.defaultBarColor = `#47DB99`;
 ```
 
-#### **description** (string)
+#### **description**: string
 Description text of the widget
 
 ```code
@@ -71,7 +78,7 @@ lang: javascript
 categoryWidget.description = 'Description';
 ```
 
-#### **heading** (string)
+#### **heading**: string
 Heading text of the widget
 
 ```code
@@ -85,7 +92,7 @@ lang: javascript
 categoryWidget.heading = 'Business Volume';
 ```
 
-#### **showClearButton** (boolean)
+#### **showClearButton**: boolean = false
 If truthy, it'll show a button to clear selected categories when there are any. Default value is `false`.
 
 ```code
@@ -94,7 +101,7 @@ lang: javascript
 categoryWidget.showClearButton = true;
 ```
 
-#### **showHeader** (boolean)
+#### **showHeader**: boolean = true
 If truthy, it'll render the heading and the component's description. Default value is `true`.
 
 ```code
@@ -103,7 +110,7 @@ lang: javascript
 categoryWidget.showHeader = false;
 ```
 
-#### **useTotalPercentage** (boolean)
+#### **useTotalPercentage**: boolean = false
 If truthy, we'll use the sum of all categories' value to render the bar percentage.
 By default, we use the maximum category value to render the bar percentage.
 
@@ -111,6 +118,15 @@ By default, we use the maximum category value to render the bar percentage.
 lang: javascript
 ---
 categoryWidget.useTotalPercentage = true;
+```
+
+#### **visibleCategories**: number = Infinity
+The number of visible categories without aggregation.
+
+```code
+lang: javascript
+---
+categoryWidget.visibleCategories = 5;
 ```
 
 ### Styles
--- a/packages/components/src/components/as-category-widget/README.md
+++ b/packages/components/src/components/as-category-widget/README.md
@@ -165,3 +165,26 @@ categoryWidget.addEventListener('categoriesSelected', event => {
   console.log('Categories Selected', event.detail)
 });
 ```
+
+### Methods
+
+#### **getSelectedCategories**
+Get current selected categories
+`Returns: Category[]`
+
+```code
+lang: javascript
+---
+const categoryWidget = document.querySelector('as-category-widget');
+categoryWidget.getSelectedCategories();
+```
+
+#### **clearSelection**
+Clear current selected categories
+
+```code
+lang: javascript
+---
+const categoryWidget = document.querySelector('as-category-widget');
+categoryWidget.clearSelection();
+```
--- a/packages/components/src/components/as-category-widget/README.md
+++ b/packages/components/src/components/as-category-widget/README.md
@@ -1,0 +1,5 @@
+# as-switch
+
+```html
+<as-switch></as-switch>
+```
--- a/packages/components/src/components/as-category-widget/README.md
+++ b/packages/components/src/components/as-category-widget/README.md
@@ -15,7 +15,7 @@ showSource: false
   description="Description"></as-category-widget>
 
 <script>
-  var categoryWidget = document.querySelector('as-category-widget');
+  const categoryWidget = document.querySelector('as-category-widget');
   categoryWidget.showClearButton = true;
   categoryWidget.categories = [
     { name: 'Bars & Restaurants', value: 1000, color: '#FABADA' },
@@ -150,4 +150,18 @@ Default: $color-ui-20 (`#F5F5F5`)
 lang: javascript
 ---
 document.body.style.setProperty('--category-bar--background-color', '#E2E6E3')
+```
+
+### Events
+
+#### **categoriesSelected**
+Fired when selected categories changed or selected categories are cleared.
+
+```code
+lang: javascript
+---
+const categoryWidget = document.querySelector('as-category-widget');
+categoryWidget.addEventListener('categoriesSelected', event => {
+  console.log('Categories Selected', event.detail)
+});
 ```
--- a/packages/components/src/components/as-category-widget/README.md
+++ b/packages/components/src/components/as-category-widget/README.md
@@ -204,3 +204,72 @@ lang: javascript
 const categoryWidget = document.querySelector('as-category-widget');
 categoryWidget.clearSelection();
 ```
+
+### Examples
+
+#### Total Percentage
+```html
+noSource: true
+---
+<iframe src="/examples/components/as-category-widget/total-percentage.html" style="width: 100%; height: 230px;">
+```
+
+```code
+lang: html
+showSource: false
+---
+<!-- Example using total sum to calculate bar percentages -->
+<as-category-widget
+  heading="Business Volume"
+  description="Description"
+  default-bar-color="#47DB99"></as-category-widget>
+
+<script>
+  const categoryWidget = document.querySelector('as-category-widget');
+  categoryWidget.useTotalPercentage = true;
+  categoryWidget.visibleCategories = 2;
+  categoryWidget.categories = [
+    { name: 'Bars & Restaurants', value: 1000, color: '#FABADA' },
+    { name: 'Fashion', value: 900 },
+    { name: 'Grocery', value: 800 },
+    { name: 'Health', value: 400 },
+    { name: 'Shopping mall', value: 250 },
+    { name: 'Transportation', value: 1000 },
+    { name: 'Leisure', value: 760 }
+  ];
+</script>
+```
+
+#### Bar only widget
+```html
+noSource: true
+---
+<iframe src="/examples/components/as-category-widget/bar-only.html" style="width: 100%; height: 256px;">
+```
+
+```code
+lang: html
+showSource: false
+---
+<!-- Example widget showing only category bars and footer -->
+<as-category-widget
+  heading="Business Volume"
+  description="Description"
+  default-bar-color="#47DB99"></as-category-widget>
+
+<script>
+  const categoryWidget = document.querySelector('as-category-widget');
+  categoryWidget.showClearButton = true;
+  categoryWidget.showHeader = false;
+  categoryWidget.visibleCategories = 5;
+  categoryWidget.categories = [
+    { name: 'Bars & Restaurants', value: 1000, color: '#FABADA' },
+    { name: 'Fashion', value: 900 },
+    { name: 'Grocery', value: 800 },
+    { name: 'Health', value: 400 },
+    { name: 'Shopping mall', value: 250 },
+    { name: 'Transportation', value: 1000 },
+    { name: 'Leisure', value: 760 }
+  ];
+</script>
+```
--- a/packages/components/src/components/as-category-widget/README.md
+++ b/packages/components/src/components/as-category-widget/README.md
@@ -1,5 +1,153 @@
-# as-switch
+Category widget displays the categories passed to the component as a list with a bar representing the percentage of each category in relation to the maximum category value.
 
 ```html
-<as-switch></as-switch>
+noSource: true
+---
+<iframe src="/examples/components/as-category-widget/simple.html" style="width: 100%; height: 330px;">
+```
+
+```code
+lang: html
+showSource: false
+---
+<as-category-widget
+  heading="Business Volume"
+  description="Description"></as-category-widget>
+
+<script>
+  var categoryWidget = document.querySelector('as-category-widget');
+  categoryWidget.showClearButton = true;
+  categoryWidget.categories = [
+    { name: 'Bars & Restaurants', value: 1000, color: '#FABADA' },
+    { name: 'Fashion', value: 900 },
+    { name: 'Grocery', value: 800 },
+    { name: 'Health', value: 400 },
+    { name: 'Shopping mall', value: 250 },
+    { name: 'Transportation', value: 1000 }
+    { name: 'Leisure', value: 760 }
+  ];
+</script>
+```
+
+### Props
+
+#### **categories** (array required)
+Array of categories, each category should include a `name` and a `value`. You can also override the bar color for each category with `color`. For example:
+
+```code
+lang: javascript
+---
+categoryWidget.categories = [
+  { name: 'A New Hope', value: 359029623, color: '#FABADA' },
+  { name: 'The Empire Strikes Back', value: 236513856 },
+  { name: 'Return of the Jedi', value: 204338075 },
+];
+```
+#### **defaultBarColor** (string)
+Overrides default color to draw the bars. Default value is `#47DB99`.
+
+```code
+lang: html
+---
+<as-category-widget defaultBarColor="#47DB99"></as-category-widget>
+```
+```code
+lang: javascript
+---
+categoryWidget.defaultBarColor = `#47DB99`;
+```
+
+#### **description** (string)
+Description text of the widget
+
+```code
+lang: html
+---
+<as-category-widget description="Description"></as-category-widget>
+```
+```code
+lang: javascript
+---
+categoryWidget.description = 'Description';
+```
+
+#### **heading** (string)
+Heading text of the widget
+
+```code
+lang: html
+---
+<as-category-widget heading="Heading"></as-category-widget>
+```
+```code
+lang: javascript
+---
+categoryWidget.heading = 'Business Volume';
+```
+
+#### **showClearButton** (boolean)
+If truthy, it'll show a button to clear selected categories when there are any. Default value is `false`.
+
+```code
+lang: javascript
+---
+categoryWidget.showClearButton = true;
+```
+
+#### **showHeader** (boolean)
+If truthy, it'll render the heading and the component's description. Default value is `true`.
+
+```code
+lang: javascript
+---
+categoryWidget.showHeader = false;
+```
+
+#### **useTotalPercentage** (boolean)
+If truthy, we'll use the sum of all categories' value to render the bar percentage.
+By default, we use the maximum category value to render the bar percentage.
+
+```code
+lang: javascript
+---
+categoryWidget.useTotalPercentage = true;
+```
+
+### Styles
+There are some CSS Variables that you can override to change visual styles.
+
+#### **--category-widget--description--color**
+Default: `#1785FB`
+
+```code
+lang: javascript
+---
+document.body.style.setProperty('--category-widget--description--color', '#1785FB')
+```
+
+#### **--category-widget--bar--height**
+Default: `4px`
+
+```code
+lang: javascript
+---
+document.body.style.setProperty('--category-widget--bar--height', '8px')
+```
+
+#### **--category-widget--background-color**
+Default: `#FFF`
+
+```code
+lang: javascript
+---
+document.body.style.setProperty('--category-widget--background-color', '#F5F5F5')
+```
+
+#### **--category-bar--background-color**
+Default: `#F5F5F5`
+
+```code
+lang: javascript
+---
+document.body.style.setProperty('--category-bar--background-color', '#E2E6E3')
 ```
--- a/packages/components/src/components/as-category-widget/as-category-widget.html
+++ b/packages/components/src/components/as-category-widget/as-category-widget.html
@@ -10,20 +10,25 @@
   </head>
 
   <body>
-    <as-category-widget></as-category-widget>
+    <as-category-widget
+      heading="Business Volume"
+      description="Description"></as-category-widget>
 
     <script>
       var categoryWidget = document.getElementsByTagName('as-category-widget')[0];
-      categoryWidget.widgetTitle = 'Business Volume';
-      categoryWidget.widgetDescription = 'Description';
+      categoryWidget.showClearButton = true;
+      categoryWidget.showHeader = true;
       categoryWidget.categories = [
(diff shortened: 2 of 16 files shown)


packages/components/src/components/as-category-widget/as-category-widget.scss
```diff
@@ -0,0 +1,81 @@
...
+
+$bar--height: get-spacing(1);
+
+as-category-widget {
+  display: block;
+
+  @include padding(4);
+
+  :root {
+    --description--color: $color-type-20;
```
IagoLast: why :root, why not declaring the css-vars under the `as-category-widget` namespace

packages/components/src/components/as-category-widget/as-category-widget.scss
```diff
@@ -0,0 +1,81 @@
...
+    --category-bar--background-color: $color-ui-20;
+    --bar--height: $bar--height;
+  }
+
+  .as-category-widget__description,
```
IagoLast: we use `&__description` in the styles, I think we should pick one and be consistent all over the project

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,195 @@
...
+    if (otherCategoryIndex >= 0) {
+      const otherCategory = newCategories.splice(otherCategoryIndex, 1)[0] as Category;
+      return { categories: newCategories, otherCategory };
+    }
+
+    return { categories: newCategories };
+  }
+}
+
+export interface Category {
```
IagoLast: should we move this interfaces to its own .ts file?

packages/components/src/components/as-category-widget/as-category-widget.scss
```diff
@@ -0,0 +1,81 @@
+@import '../../../../styles/src/core/_dev';
+
+$bar--height: get-spacing(1);
+
+as-category-widget {
+  display: block;
+
+  @include padding(4);
```
IagoLast: do we want to always have padding or do we want the container element to define this padding?

imagine we are placing this inside a `floating-panel` who should be the responsible of this padding? the category widget or the floating panel should define a padding for its content?

packages/components/src/components/as-category-widget/as-category-widget.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+    });
+
+    it('should clear selected categories', () => {
+      const selectedCategories = ['Category 1'];
+      categoryWidget.selectedCategories = selectedCategories;
+
+      categoryWidget.clearSelection();
+
+      expect(categoryWidget.getSelectedCategories()).toEqual([]);
+      expect(categoryWidget._onCategoriesChanged).toHaveBeenCalled();
```
IagoLast: I think we want to test that the component fires some kind of event when the categories are cleared. I think with this expect, we are [testing the implementation](https://teamgaslight.com/blog/testing-behavior-vs-testing-implementation) but not the desired behavior.

I think we could create 2 new tests for .clearSelection() checking the component should trigger an event with previously selected categories and...with no selected categories should the component trigger an event? 🤔

packages/components/src/components/as-category-widget/as-category-widget.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  describe('._toggleCategory', () => {
+    beforeEach(() => {
+      spyOn(categoryWidget, '_onCategoriesChanged');
+    });
+
+    it('should select category if it is not selected', () => {
+      const category = { name: 'Category 1', value: 10, color: '#000' };
+      categoryWidget._toggleCategory(category);
+      expect(categoryWidget._onCategoriesChanged).toHaveBeenCalled();
```
IagoLast: similar issue here, I think testing the real event would result in a more strong test, this way if we change the implementation (ie, renaming _onCategoriesChanged to _onCategoriesChange) the test will fail

packages/components/src/components/as-category-widget/as-category-widget.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+        components: [CategoryWidget],
+        html: '<as-category-widget></as-category-widget>'
+      });
+    });
+
+    it('should emit an event containing selected categories', async () => {
+      element.categories = exampleCategories;
+      await testWindow.flush();
+
+      const spy = jest.fn();
```
IagoLast: I think if we rename this to categoriesSelectedSpy the test is easier to read

packages/components/src/components/as-category-widget/as-category-widget.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+
+      expect(spy).toHaveBeenCalled();
+      expect(spy.mock.calls[0][0].detail).toEqual([exampleCategories[0].name]);
+    });
+  });
+
+  describe('._getVisibleCategoriesMaximumValue', () => {
+    it('should return maximum value of visible categories', () => {
+      categoryWidget.categories = exampleCategories;
+      categoryWidget.useTotalPercentage = false;
```
IagoLast: what is `useTotalPercentage` doing?

packages/components/src/components/as-category-widget/as-category-widget.scss
jesusbotella: Yes, it would be better doing it that way.

packages/components/src/components/as-category-widget/as-category-widget.scss
jesusbotella: Yes, I think so but the thing here is that it is not `.as-category-widget` but `as-category-widget`. So whether we wrap these styles in `.as-category-widget` or we set that class to the host element and transform `as-category-widget` to `.as-category-widget`.

What do you prefer?

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: I don't know, maybe we can move them to separated files if we use them anywhere else. What do you think?

packages/components/src/components/as-category-widget/as-category-widget.scss
IagoLast: oooouch, you are right... so its fine this way

packages/components/src/components/as-category-widget/as-category-widget.tsx
IagoLast: As a former java user Im used to have classes/Interfaces splitted in different files but let @rubenmoya @ivanmalagon give their opinion

packages/components/src/components/as-category-widget/as-category-widget.scss
jesusbotella: Yes, that's a matter to discuss, because if you want to use it as a standalone component the padding would be needed, right? 🤷‍♂️

IagoLast: right! IMO the component itself should not define any padding/margin...etc IMO that's responsibility of the container/user

packages/components/src/components/as-category-widget/as-category-widget.spec.ts
jesusbotella: It should always trigger an event when categories are updated, to keep the user updated about the state of the widget.

I only need to modify the current test to listen to the fired event.

As I am testing a similar thing in `._onCategoriesChanged` describe section, I am going to rename that section to `Behaviour` and place that test there.

jesusbotella: No, that's not expected, I thought you meant other thing 🤦‍♂️

packages/components/src/components/as-category-widget/as-category-widget.html
```diff
@@ -0,0 +1,33 @@
...
+    <script src="/build/airship.js"></script>
+  </head>
+
+  <body>
+    <as-category-widget
+      heading="Business Volume"
+      description="Description"></as-category-widget>
+
+    <script>
+      var categoryWidget = document.getElementsByTagName('as-category-widget')[0];
```
IagoLast:
```html
var categoryWidget = document.querySelector('as-category-widget');
```

also do the job

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,195 @@
...
+  private _getBarColor(color: string, options: { isSelected?: boolean } = {}) {
+    if (options.isSelected) {
+      return shadeOrBlend(-0.16, color);
+    }
+
+    return color;
+  }
+
+  private parseCategories() {
+    const newCategories = [...this.categories];
```
IagoLast: Why not using [find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)

If stencill dont do it, we can load a polyfill for IE  I think having a clean code it worth it ;D 

```ts
private parseCategories() {
    const categories = [...this.categories];
    const otherCategory = this.categories.find((category: Category) => category.name === OTHER_CATEGORY_NAME);
    if (otherCategory) {
      return { categories, otherCategory };
    }
    return { categories };
  }
```

IagoLast: do we need the spread operator here?

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  private _getVisibleCategoriesMaximumValue() {
+    const categories = this.categories.slice(0, this.numberOfVisibleCategories);
+
+    return categories.reduce(
+      (maximum, currentCategory: Category) => currentCategory.value > maximum ? currentCategory.value : maximum, 0
+    );
+  }
+
+  private _getCategoriesTotalValue(categories: object[]) {
```
IagoLast: Isn't categories a Category array?

```ts
private _getCategoriesTotalValue(categories: Category[]) {}
```

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,195 @@
...
+    return categories.reduce(
+      (maximum, currentCategory: Category) => currentCategory.value + maximum, 0
```
IagoLast: This is not the maximun, is the `accumulator` or the `acum`... copy&paste?  👀

IagoLast: we could also specify type `reduce(acum: number, currentCategory: Category) ...`

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,195 @@
...
+    this.selectedCategories = newSelectedCategories;
+    this._onCategoriesChanged();
+  }
+
+  private _onCategoriesChanged() {
+    this.categoriesSelected.emit(this.selectedCategories);
+  }
+
+  private _getVisibleCategoriesMaximumValue() {
+    const categories = this.categories.slice(0, this.numberOfVisibleCategories);
```
IagoLast: we could create a function 

```ts
private _getVisibleCategories() {
  return this.categories.slice(0, this.numberOfVisibleCategories);
}
```

Then we could use this funcion here 

```ts
private _getVisibleCategoriesMaximumValue() {
    const visibleCategories = this._getVisibleCategories();

    return visibleCategories.reduce((maximum, currentCategory: Category) => {
      return currentCategory.value > maximum ? currentCategory.value : maximum;
    }, 0);
  }
```

or even

```ts
private _getVisibleCategoriesMaximumValue() {
    return this._getVisibleCategories().reduce((maximum, currentCategory: Category) => {
      return currentCategory.value > maximum ? currentCategory.value : maximum;
    }, 0);
  }
```

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,195 @@
...
+export class CategoryWidget {
+  @Prop() public categories: object[] = [];
+  @Prop() public defaultBarColor: string = DEFAULT_BAR_COLOR;
+  @Prop() public description: string;
+  @Prop() public heading: string;
+  @Prop() public showClearButton: boolean;
+  @Prop() public showHeader: boolean = true;
+  @Prop() public useTotalPercentage: boolean = false;
+
+  @Event() public categoriesSelected: EventEmitter<string[]>;
```
IagoLast: we could give a prefix/subfix to eventEmitters in components to easily detect them

`categoriesSelectedEmitter`?

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+        <div class='as-category-widget__bar'>
+          <div class='as-category-widget__value' style={progressStyles}></div>
+        </div>
+      </li>
+    );
+  }
+
+  private _renderOtherCategory(category: Category, options: CategoryOptions) {
+    const categoryData = category || {
```
IagoLast: since typescript supports default parameters we could use them, but the syntax is not very familiar in this case:

```ts
private _renderOtherCategory(
    category: Category = {
      name: 'Other',
      value: this._getCategoriesTotalValue(
        this.categories.slice(this.numberOfVisibleCategories, this.categories.length)
      )
    },
    options: CategoryOptions) {

    return this._renderCategory(category, { maximumValue: options.maximumValue, isOther: true });
  }
```

jesusbotella: No, I think it's easier to follow as it is right now. That syntax looks weird.

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: I don't like accumulator at all, and even less being shortened. I am gonna put `currentMaximum`.

We don't need to be explicit here, TypeScript infers parameter's type.

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: Yes, it is but as you can see `categories` in the class definition is `object[]`, and needs to be like that for props definition. So I think it's easier to be like like that, do you agree?

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: Yes, we need it because we're cloning the array, `[this.categories]` wouldn't work.

packages/components/src/components/as-category-widget/as-category-widget.html
jesusbotella: 👍

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: I thought about it first but I don't know why I discarded it 🤔. Definitely seems cleaner.

jesusbotella: I have just realized that we can't do that, because I want `otherCategory` to be out of the array.

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: 👍

jesusbotella: We can't name it like that because that variable name is used for the name of the event, so it would be `addEventListener('categoriesSelectedEmitter')`

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: oh sh*t, yeah, I know what you mean 😂. What do you think about `sum`?

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    this.selectedCategories = [];
+    this._onCategoriesChanged();
+  }
+
+  public render() {
+    return [
+      this._renderHeader(),
+      <ul class='as-category-widget__list'>{this._renderCategories()}</ul>,
```
IagoLast: Is this too much? I'd love to have 3 equal methods 
```jsx
private function _renderContent() {
  return <ul class='as-category-widget__list'>{this._renderCategories()}</ul>
}
```

jesusbotella: No, I thought about it too, would be good :)

I feel like I did a poor function ordering in the class, I think it is convoluted somehow 🤷‍♂️

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+    const categoriesToRender =  moreCategoriesThanVisible
+      ? categories.slice(0, this.numberOfVisibleCategories)
+      : categories;
+
+    const maximumValue = this.useTotalPercentage
+      ? this._getCategoriesTotalValue(this.categories)
+      : this._getVisibleCategoriesMaximumValue();
+
+    if (otherCategory || moreCategoriesThanVisible) {
```
IagoLast: wow the logic is not very straightforward here, I wasn't expecting `_renderOtherCategory` to render  a default `others` when `otherCategory` parameter is not given... not sure about the best solution

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,203 @@
...
+    };
+
+    return this._renderCategory(categoryData, { maximumValue: options.maximumValue, isOther: true });
+  }
+
+  private _renderFooter() {
+    const selectedCount = this.selectedCategories.length ? this.selectedCategories.length : 'All';
+
+    return (
+      <footer class='as-category-widget__footer'>
```
IagoLast: not sure if footer is the right tag for this, is not an obligation but a recommendation, see [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/footer), [this](https://www.w3.org/wiki/Html/Elements/footer) or [this](https://www.w3schools.com/tags/tag_footer.asp) 

> The footer tag defines a footer for a document or section. A footer element should contain information about its containing element. typically contains: authorship information, copyright information, contact information, sitemap. back to top links, related documents

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: I dunno either 🤷‍♂️

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,203 @@
...
+        <div class='as-category-widget__count as-body'>{selectedCount} selected</div>
+        { this.showClearButton
+          ? <button class='as-category-widget__clear' onClick={() => this.clearSelection()}>Clear selection</button>
+          : '' }
+      </footer>
+    );
+  }
+
+  private _isSelected(categoryName: string) {
+    return this.selectedCategories.indexOf(categoryName) > -1;
```
IagoLast: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

?

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  private _toggleCategory(category: Category) {
+    const newSelectedCategories = this._isSelected(category.name)
```
IagoLast: can we assign directly the result of the ternary operator to this.selectedCategories?

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+  private _getBarColor(color: string, options: { isSelected?: boolean } = {}) {
+    if (options.isSelected) {
+      return shadeOrBlend(-0.16, color);
+    }
+
+    return color;
+  }
+
+  private parseCategories() {
```
IagoLast: _parseCategories?

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,203 @@
...
+      value: this._getCategoriesTotalValue(
+        this.categories.slice(this.numberOfVisibleCategories, this.categories.length)
+      )
+    };
+
+    return this._renderCategory(categoryData, { maximumValue: options.maximumValue, isOther: true });
+  }
+
+  private _renderFooter() {
+    const selectedCount = this.selectedCategories.length ? this.selectedCategories.length : 'All';
```
IagoLast: `const selectedCount = this.selectedCategories.length || 'All'`

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,203 @@
...
+    return [
+      categoriesToRender.map((category: Category) => this._renderCategory(category, { maximumValue })),
+      otherCategoryTemplate
+    ];
+  }
+
+  private _renderCategory(category: Category, options: CategoryOptions) {
+    const isSelected = this._isSelected(category.name);
+    const isAnyCategorySelected = this.selectedCategories.length > 0;
+    const barColor = !options.isOther
```
IagoLast: why comparing against the negated condition? 😂  and not just:
```js
const barColor = options.isOther
      ? OTHER_CATEGORY_COLOR
      : this._getBarColor(category.color || this.defaultBarColor, { isSelected });
```

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: I think it contains information about its containing element, right?
It has the information of how many categories are selected

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: Because I found it more intuitive to put the `_getBarColor` in the upper branch but I don't mind :)

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,203 @@
...
+  @Prop() public description: string;
+  @Prop() public heading: string;
+  @Prop() public showClearButton: boolean;
+  @Prop() public showHeader: boolean = true;
+  @Prop() public useTotalPercentage: boolean = false;
+
+  @Event() public categoriesSelected: EventEmitter<string[]>;
+
+  @State() private selectedCategories: string[] = [];
+  @State() private numberOfVisibleCategories: number = 5;
```
VictorVelarde: IMO this should be easily set from outside by the client

jesusbotella:
I think I have made a bad decision when it comes to showing categories, I should have allowed to show all categories by default. I am going to refactor the logic to allow that.

packages/components/src/components/as-category-widget/README.md
```diff
@@ -0,0 +1,190 @@
...
+<iframe src="/examples/components/as-category-widget/simple.html" style="width: 100%; height: 330px;">
+```
+
+```code
+lang: html
+showSource: false
+---
+<as-category-widget
+  heading="Business Volume"
+  description="Description"></as-category-widget>
```
IagoLast: we could pass all possible parameters here with the default values

packages/components/src/components/as-category-widget/README.md
```diff
@@ -0,0 +1,190 @@
...
+    { name: 'Shopping mall', value: 250 },
+    { name: 'Transportation', value: 1000 }
+    { name: 'Leisure', value: 760 }
+  ];
+</script>
+```
+
+### Props
+
+#### **categories** (array required)
```
IagoLast: I think the most common way to document APIS in markdown is:

Required parameter:
name:type

Optional parameter:
[name:type]=default


cc @ivanmalagon @rubenmoya

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: I added a property named `visibleCategories` to do that :) Thank you!

packages/components/src/components/as-category-widget/as-category-widget.scss
```diff
@@ -0,0 +1,81 @@
...
+  .as-category-widget__bar {
+    position: relative;
+    width: 100%;
+    height: $bar--height;
+    border-radius: get-spacing(0.5);
+    background-color: var(--category-bar--background-color, $color-ui-20);
+    background-color: $color-ui-20;
+  }
+
+  .as-category-widget__value {
```
piensaenpixel: What do you think if we avoid nesting the classes? i know it's more organized ... but I'd like a clearer css output

packages/components/src/components/as-category-widget/as-category-widget.scss
```diff
@@ -0,0 +1,81 @@
...
+
+  .as-category-widget__description + .as-category-widget__list {
+    margin-top: get-spacing(4);
+  }
+
+  .as-category-widget__category-title {
+    position: relative;
+
+    &::after {
+      content: attr(data-value);
```
piensaenpixel: I'm not sure about using the pseudoelement for showing content in the html

jesusbotella: Yes, I didn't think about accessibility 😞.
And leveraging this situation I think I am going to make an improvement and use an ellipsis to cut overflowed text :)

packages/components/src/components/as-category-widget/as-category-widget.scss
jesusbotella: Nesting inside `as-category-widget`? I don't really mind putting them outside `as-category-widget`. We're using BEM so there shouldn't be any collisions :)

What do you think @IagoLast @rubenmoya @ivanmalagon?

jesusbotella:
I changed @piensaenpixel suggestions and added an ellipsis to category title, so that when there's more text than maximum width there are `...` cutting it.

screen shot 2018-08-01 at 13 10 47

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,210 @@
...
+  private _renderCategoryList() {
+    return <ul class='as-category-widget__list'>{this._renderCategories()}</ul>;
+  }
+
+  private _renderCategories() {
+    const moreCategoriesThanVisible = this.categories.length > this.visibleCategories;
+    const { categories, otherCategory } = this._parseCategories();
+    let otherCategoryTemplate;
+
+    const categoriesToRender =  moreCategoriesThanVisible
```
rubenmoya: I think we can avoid this ternary you can always do `categories.slice(0, this.visibleCategories)`, if the array is smaller it will return it with the same size, in fact I think we can just use `this._getVisibleCategories`

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,210 @@
...
+    return this._renderCategory(categoryData, { maximumValue: options.maximumValue, isOther: true });
+  }
+
+  private _renderFooter() {
+    const selectedCount = this.selectedCategories.length || 'All';
+
+    return (
+      <footer class='as-category-widget__footer'>
+        <div class='as-category-widget__count as-body'>{selectedCount} selected</div>
+        { this.showClearButton
```
rubenmoya: What I use to do in react was:

```
{ this.showClearButton && <button class='as-category-widget__clear' onClick={() => this.clearSelection()}>Clear selection</button> }
```

I think it looks cleaner, but it's just my opinion 😂

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -0,0 +1,210 @@
...
+    return [
+      categoriesToRender.map((category: Category) => this._renderCategory(category, { maximumValue })),
+      otherCategoryTemplate
+    ];
+  }
+
+  private _renderCategory(category: Category, options: CategoryOptions) {
+    const isSelected = this._isSelected(category.name);
+    const isAnyCategorySelected = this.selectedCategories.length > 0;
+    const barColor = options.isOther
```
rubenmoya: Can't we move the `isOther` logic to the `getBarColor` method? In my head the `renderCategory`method shouldn't do this work, I expect `getBarColor` to be smart enough to give me the final color, also, we are doing something similar with the `isSelected` what do you think?

packages/components/src/utils/styles.js
```diff
@@ -0,0 +1,25 @@
+
+// pSBC.js
+// From Pimped JavaScript (https://github.com/PimpTrizkit/PJs)
+// Usage: https://github.com/PimpTrizkit/PJs/wiki/12.-Shade,-Blend-and-Convert-a-Web-Color-(pSBC.js)
+export function shadeOrBlend(p, from, to) {
```
rubenmoya: <a href="https://imgflip.com/i/2f59e6"></a>

packages/components/src/components/as-category-widget/as-category-widget.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+    });
+
+    it('should render properly', async () => {
+      element.heading = 'Category Widget Example';
+      element.description = 'Description for Category Widget';
+      element.categories = exampleCategories;
+      element.showClearButton = true;
+      await testWindow.flush();
+
+      expect(element).toMatchSnapshot();
```
ivanmalagon: What is the snapshot that is compared to? Where?

IagoLast: https://jestjs.io/docs/en/snapshot-testing

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: Yes, it would be so much better :)

packages/components/src/components/as-category-widget/as-category-widget.spec.ts
jesusbotella: That's basically it. It creates the references in a folder, which is commited to the repo, to be able to compare them in every test run. If there's something different, it fails.

jesusbotella:
I have just added JSDocs that I completely forgot about 😂, and added two more examples at the bottom. One using total sum percentage calculation, and other showing only category bars.

PR #109, opened 2018-07-31
Add simple postcss plugin

IagoLast:
This PR adds a example of how to use postcss from npm-scripts without installing any external a task runner.

```bash
npm run postcss
```


IagoLast:
🚫 closed, we will address this in the future

Issue #111, opened 2018-08-01
Remove CSS reset


rubenmoya:
### Context
We're using a CSS reset to reset all element styles like margins, paddings, etc, this could be good for us as developers, but since those styles are affecting the HTML elements directly (without classes) we could be overriding people styles.

Also, by setting every margin and padding to 0 we're breaking the whole application spacing.

We should remove the reset file and reset only the needed styles in each element/component, that way we won't mess with the user's styles.

PR #112, opened 2018-08-01
Remove reset.css

rubenmoya:
#111 Remove reset.css
---


packages/styles/src/core/lib/_reset.scss
```diff
@@ -1,155 +0,0 @@
...
-  content: '';
-  content: none;
-}
-
-table {
-  border-spacing: 0;
-  border-collapse: collapse;
-}
-
-main {
```
IagoLast: This is needed to IE11

packages/styles/src/core/lib/_reset.scss
```diff
@@ -1,155 +0,0 @@
...
-  display: block;
-}
-
-button {
```
IagoLast: This is needed for the tabs to look ✨

packages/styles/src/core/lib/_reset.scss
```diff
@@ -1,155 +0,0 @@
-/* http://meyerweb.com/eric/tools/css/reset/
-   v2.0 | 20110126
-   License: none (public domain)
-*/
-
-/* apply a natural box layout model to all elements, but allowing components to change */
-html {
-  box-sizing: border-box;
```
IagoLast: This is needed of everything is broken

packages/styles/src/core/lib/_reset.scss
rubenmoya: This is added in the .tabs__item class

packages/styles/src/core/lib/_reset.scss
rubenmoya: To be honest I'm not sure what to do in this case, since it's just for IE11 we could add it to our `core.css`

packages/styles/src/core/lib/_reset.scss
rubenmoya: Also, I just realized we're adding it in `core.css`:

```
* {
  box-sizing: border-box;
}
```

packages/styles/src/core/lib/_reset.scss
IagoLast: 👍

packages/styles/src/core/lib/_reset.scss
IagoLast: 👍

packages/styles/src/core/lib/_reset.scss
IagoLast: yep, perfect

Issue #43, opened 2018-07-16
Elements CSS


ivanmalagon:
We have to implement the basic elements of the Airship framework. Please use as reference https://projects.invisionapp.com/share/3REYL6RJK

These elements will be built using only CSS. For most complex elements we'll create Custom Elements.

If you find that any them that can't be built only with CSS, please give the team a heads up.

Basic CSS
- [x] Colors #75
- [x] Typography #76
  - [x] Jumbo
  - [x] Display
  - [x] Title (normal / bold)
  - [x] Subheader (normal / bold / mono)
  - [x] Body (normal / bold / mono)
  - [x] Caption (normal / bold / mono)
- [x] List #76
  - [x] Bulleted
  - [x] Numbered
- [x] Icons #77

Elements
- [x] Buttons
- [x] Inputs, textarea
- [x] Checkbox
- [x] Radio buttons
- [x] Tabs (WC?)
- [x] Banner
- [x] Flags
- [x] Avatar
- [x] Tooltips
- [x] Breadcrumbs
- [x] Menu dropdown (perhaps we can reuse dropdown style. Create a WC as well?)
- [x] Badges
- [x] Switch (should we create a WC also?)
- [x] Loading (global, block, title)

Nice to have
- [x] Table

PR #115, opened 2018-08-01
Add inputs

rubenmoya:
#43
----

- [x] Implementation
- [x] Design Aproval
- [x] UI Test
- [x] CR


packages/styles/src/inputs/_inputs.scss
```diff
@@ -0,0 +1,110 @@
...
+  min-height: 32px;
+  padding: 6px 8px;
+  border: 0;
+  border-radius: 4px;
+  background: var(--as-color-ui-01, #{$color-ui-01});
+  box-shadow: inset 0 0 0 1px var(--as-color-ui-04, #{$color-ui-04});
+  color: var(--as-color-type-01, #{$color-type-01});
+  font: $font-body;
+
+  &::-webkit-input-placeholder {
```
IagoLast: can we delegate this browser-specific selectors in a postcss plugin?

packages/styles/src/inputs/_inputs.scss
```diff
@@ -0,0 +1,110 @@
+@import "../core/_dev";
+
+.as-input,
+.as-textarea {
+  box-sizing: border-box;
+  width: 100%;
+  min-height: 32px;
+  padding: 6px 8px;
```
IagoLast: should we use relative rules?

packages/styles/src/inputs/_inputs.scss
rubenmoya: Great idea! It seems like we can use `::placeholder` and it takes care of it :D

screen shot 2018-08-02 at 13 04 45

piensaenpixel:
Could we add a transition (as we have in the radio buttons) in the focus state?

piensaenpixel:
The rest is perfect!

PR #116, opened 2018-08-01
43 buttons

IagoLast:
#43 
---

- [x] Implementation
- [x] Design Aproval
- [x] UI Test
- [ ] CR


rubenmoya:
- The buttons should have smooth a transition between states
- The disabled button should have the `not-allowed` cursor
- The disabled button should have the `pointer-events: none` rule
- In safari the button as a margin, which doesn't have in chrome

IagoLast:
@rubenmoya PR updated 👍

packages/styles/src/buttons/buttons.scss
```diff
@@ -60,3 +60,35 @@
...
+  }
+
+  &:active {
+    background-color: var(--as-button-color-focus, #{$color-ui-03});
+  }
+}
+
+a.as-btn.as-btn--primary {
+  &:hover {
+    background-color: var(--as-button-primary-color-focus, darken($color-primary, 20%));
```
rubenmoya: Do we need to interpolate this too?

packages/styles/src/buttons/buttons.scss
```diff
@@ -60,3 +60,35 @@
     font: $font-subheader;
   }
 }
+
+a.as-btn {
+  text-decoration: none;
+
+  &:hover {
+    background-color: var(--as-button-color-focus, #{$color-ui-02});
```
rubenmoya: This is something I want to talk later in the retrospective meeting, should we define these variables on top of the component so we know these variables are created here?

piensaenpixel:
Hi, some things

- Focus state, we should remove the outline
image

- Border should be '0' instead of 1px transparent
image

packages/styles/src/buttons/buttons.scss
IagoLast: yes, sure

packages/styles/src/buttons/buttons.scss
IagoLast: nope

IagoLast:
@piensaenpixel 

> Focus state, we should remove the outline

how do we give focus feedback then?

> Border should be '0' instead of 1px transparent

So secondary button will be bigger because of its border, is the expected behaviour?

piensaenpixel:
- Focus should have the same design that the press state
- we need use the border-size property to avoid the different size.

piensaenpixel:
BTW now the button is 42px instead 40, remember all the sizes should be multiple of 4

piensaenpixel:
-  Here we have a double border
image

-  And i think the markup should be a nav or at least a list

The rest is ok for me

PR #121, opened 2018-08-02
Banner

rubenmoya:
#43
----

- [x] Implementation
- [x] Design Aproval
- [ ] UI Test
- [x] CR


packages/styles/src/banner/_banner.scss
```diff
@@ -0,0 +1,32 @@
...
+    flex: 1;
+    padding: 0 8px;
+  }
+
+  &__icon {
+    display: flex;
+    flex: 0 0 auto;
+  }
+
+  &--support-01 {
```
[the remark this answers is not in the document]
rubenmoya: Yep, my first thought was to use `--success`, `--warning` and `--error`, but since we used the color names (`--type-01`) in the typography modifiers I decided to keep it that way.

In the case of support colors it's easy, you have `success`, `warning`, and `error`, but what about the typography? I can't think of 4 names for them, maximum 2 😂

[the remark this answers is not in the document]
IagoLast: we could greate semantic aliases

--support-02 == warning
--type-04 == gray

packages/styles/src/banner/README.md
```diff
@@ -0,0 +1,45 @@
...
+  </div>
+  <div class="as-banner__content">Your profile settings have been updated successfully</div>
+</div>
+```
+
+## For warnings
+
+```html
+<div class="as-banner as-bg--warning as-color--type-04">
+  <div class="as-banner__icon">
```
IagoLast: what happens if we want to use a `img` ?

piensaenpixel:
like a charm

packages/styles/src/airship.scss
```diff
@@ -1,7 +1,9 @@
 @import './core/core.scss';
 @import './avatar/avatar.scss';
+@import './banner/banner.scss';
 @import './breadcrumb/breadcrumb.scss';
 @import './inputs/inputs.scss';
 @import './flag/flag.scss';
 @import './tabs/tabs.scss';
 @import './radio/radio.scss';
+@import './utilities/_colors';
```
IagoLast: create a utilities file and import everything inside utilities folder there

packages/styles/src/banner/README.md
rubenmoya: you can use whatever you want, inside that div

packages/styles/src/airship.scss
```diff
@@ -1,7 +1,9 @@
 @import './core/core.scss';
 @import './avatar/avatar.scss';
+@import './banner/banner.scss';
 @import './breadcrumb/breadcrumb.scss';
 @import './inputs/inputs.scss';
 @import './flag/flag.scss';
 @import './tabs/tabs.scss';
 @import './radio/radio.scss';
+@import './utilities/_utilities';
```
IagoLast: utilities should not be prefixed so it can be used in a separated way

rubenmoya: cmon man, let me merge this already!!!!! 😂

PR #122, opened 2018-08-02
Flags

rubenmoya:
#43
----

- [x] Implementation
- [x] Design Aproval
- [x] UI Test
- [x] CR


packages/styles/src/flags/README.md
```diff
@@ -0,0 +1,46 @@
+# Flag
+
+Flag displays contextual information about events on the page.
+
+## Default
+
+```html
+<div class="as-flag">
+  <div class="as-flag__icon">
+    <svg width="12px" height="13px" viewBox="0 0 16 15" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
IagoLast: IMO using Inline svgs in the examples adds a lot of noise

packages/styles/src/flags/_flags.scss
```diff
@@ -0,0 +1,29 @@
+.as-flag {
+  display: flex;
+  max-width: 260px;
+  padding: 16px;
```
IagoLast: sizes should be relative to our base unit

packages/styles/src/flags/README.md
rubenmoya: Yep, I know, we can change it later when we have the `<as-icon />` component

piensaenpixel:
some comments:

- I think it could be great to have different sizes: 260 / 320 / 400 and 100%
- I'd try to add a bit more semantic markup in titles and body text
- Close icon should be 12px instead 14 and it should have a cursor pointer
image

packages/styles/src/flags/README.md
IagoLast: icons are ready now, you can use the svg version here ^^

https://carto-airship.netlify.com/catalog/#/icons/icons

piensaenpixel:
Like a charm

PR #123, opened 2018-08-02
Breadcrumbs

rubenmoya:
#43
----

- [x] Implementation
- [x] Design Aproval
- [x] UI Test
- [x] CR


packages/styles/src/breadcrumb/_breadcrumb.scss
```diff
@@ -0,0 +1,34 @@
...
+      color: var(--as-color-primary, #{$color-primary});
+      font-size: 12px;
+    }
+
+    &:last-child::after {
+      display: none;
+    }
+  }
+
+  &__link {
```
IagoLast: I think we could just overwrite the `a` element styles so the user has to write less code

PR #139, opened 2018-08-03
Update test suite

IagoLast:
This PR makes writing a test easier. Now we just need the name of the `HTML` file to be tested and our test-utils manage the rest.


packages/styles/src/core/containers/test/containers.spec.js
```diff
@@ -1,52 +1,14 @@
...
-  u.spec(sidebarMobileReference, sidebarMobileScreenshot, sidebarUrl, true),
-  u.spec(bottomDesktopReference, bottomDesktopScreenshot, bottomUrl),
-  u.spec(bottomMobileReference, bottomMobileScreenshot, bottomUrl, true),
-  u.spec(legendsDesktopReference, legendsDesktopScreenshot, legendsUrl),
-  u.spec(legendsMobileReference, legendsMobileScreenshot, legendsUrl, true),
-  u.spec(allDesktopReference, allDesktopScreenshot, allUrl)
+  utils.spec('containers-in-sidebar'),
+  utils.spec('containers-in-sidebar', true),
+  utils.spec('containers-in-bottom'),
+  utils.spec('containers-in-bottom', true),
```
rubenmoya: I'd use an options object here to give more context, right now we don't know what `true` means

IagoLast:
> It really looks much better, although we still having the problem of not being able to know what changed, or what's broken which is the biggest problem in my opinion

Agreed, but I think we should let that for another PR
@rubenmoya

rubenmoya:
Yes, for sure, but I'd change the `true` to be an options object :P

Issue #94, opened 2018-07-25
Range slider widget


ivanmalagon:
A component that allows users to input numeric values within a range by dragging its sliders.

![image](https://user-images.githubusercontent.com/1078228/43202004-7d33bcb2-901a-11e8-844c-f67a19169570.png)

The React component done in our first iteration is great so please refer also to [its documentation](https://carto-airship.netlify.com/#/components/range).

Inspired by the React component, I'll proceed to copypaste here the API.

### Props

**maxValue**
Number.
The top limit of the range. You cannot drag your slider beyond this value. By default the value is 10.

**minValue**
Number
The bottom limit of the range. You cannot drag your slider under this value. By default the value is 0.

**value**
Number
The initial value. By default the value is 0 or the minValue.

**disabled**
Boolean
To get the component disabled.

**name**
String
In order to make the component accesible, it creates hidden inputs underneath when you pass a name property.

**step**
Number
The default increment/decrement of your component is 1. You can change that by setting a different number to this property.

**draggable**
Boolean
If this property is set to true, and it has multiple value, you can drag the entire track.

**formatLabel**
Number
By default, value labels are displayed as plain numbers. If you want to change the display, you can do so by passing in a function.

### Callbacks

**changeStart**
Triggered when the user starts dragging a slider. Return the current range.

**change**
Triggered when the user is dragging a slider. Return the current range.

**changeEnd**
Triggered when the user ends dragging a slider. Return the current range.

### Questions
- I miss in the React docs, how can I set the slider to one or two draggable balls :raur: ? @rubenmoya

jesusbotella:
Done in #141!

PR #141, opened 2018-08-06
Range slider

ivanmalagon:
#94 
---
6-Aug-2018

Copypasting from @jesusbotella comment

---

I have just pushed the latest changes of RangeSlider. There are 2 things missing, setting aria attributes (aria-valuemax, ariavaluemin…) to <as-range-slider> tag and apply step transforming function to slider values (you can find an example function in React’s slider transformer.js file). Everything apart from that should be working OK.

There’s some logic which can be extracted to auxiliar methods, and SCSS might need some variables over there, but it should be OK :slightly_smiling_face:

---

https://deploy-preview-141--carto-airship.netlify.com/catalog/#/components/as-range-slider


VictorVelarde:
✨ Everything LGTM in your latest changes @jesusbotella 
Great improvement in bar movement and specs!

examples/layouts/basic-layout.html
```diff
@@ -44,4 +44,4 @@
   <script defer src="./common/map.js"></script>
 </body>
 
-</html>
+</html>
```
IagoLast: 👀

IagoLast: same issue multiple files

jesusbotella: I think @VictorVelarde has a preset to remove trailing blank lines from files :(

I am gonna revert those changes.

VictorVelarde: Sorry, I'll check it out. My user settings includes `"files.trimTrailingWhitespace": false` but maybe it is some other extension

VictorVelarde: I've changed my VSCode user settings to this:
```
 "files.trimTrailingWhitespace": true,
 "files.insertFinalNewline": true,
 "files.trimFinalNewlines": true,
```
This guarantees one, and only one, new line at EOF

PR #142, opened 2018-08-06
Research BackstopJS as visual regression testing

ivanmalagon:


DEVELOPERS.md
```diff
@@ -4,17 +4,44 @@
...
+- Create a `test` folder within the component folder.
+- Create as many `html` as needed to check every possible state of the component.
+- Create a `wadus.spec.js` that returns an array of specs consisting of the test label and the local `.html` file that asserts that scenario. The format is:
+
+```
+const Utils = require('../../test-utils');
+const u = new Utils(__dirname);
+
+module.exports = [
+  u.spec('Tabs', 'tabs.html'),
```
IagoLast: we can use same name for both and save some coding here

.circleci/config.yml
```diff
@@ -42,21 +42,20 @@ jobs:
...
-      - run: Xvfb :98 &
       - restore_cache:
           keys:
           - v1-dependencies-{{ checksum "package.json" }}
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
       - run: npm install
       - run: npm run build:styles
-      - run: xvfb-run -a npm run test:styles
+      - run: npm run ci:styles
```
IagoLast: minor thing but I prefer `test:ci` since all our test related tasks start with `test:`

.gitignore
```diff
@@ -24,3 +24,6 @@ node_modules/
 Thumbs.db
 UserInterfaceState.xcuserstate
 www/
+backstop_data/bitmaps_test/
+backstop_data/ci_report/
+backstop_data/html_report/
```
IagoLast: missing newline

DEVELOPERS.md
```diff
@@ -4,17 +4,44 @@
...
+  u.spec('Tabs XL', 'tabs-xl.html'),
+  u.spec('Tabs accesible', 'tabs-accesible.html')
+];
+```
+
+- Run `npm run visual:reference`. This will create the new references. Check them and, if they're OK, commit them to the project. `BackstopJS` will use these references from now on.
+
+**Recreate image references**
+
+If you make any change in an existing visual spec, run `npm run visual:reference`. This command recreates the reference and if there's any difference, the current git diff will tell you that there are changes in an existing reference.
```
IagoLast: npm run `test:prepare`   ?

backstop.config.js
```diff
@@ -0,0 +1,64 @@
...
+    "args": ["--no-sandbox"]
+  },
+  "asyncCaptureLimit": 5,
+  "asyncCompareLimit": 50,
+  "debug": false,
+  "debugWindow": false
+};
+
+const report = {
+  report: argv.report === 'ci'
```
IagoLast: we can use the env vars instead arguments

backstop.config.js
```diff
@@ -0,0 +1,64 @@
...
+const path = require('path');
+const argv = require('minimist')(process.argv.slice(2));
+
+const defaultScenarioOptions = {
+  misMatchThreshold: 0,
+  requireSameDimensions: true,
+  delay: 100
+};
+
+let defaultOptions = {
```
IagoLast: const ?

IagoLast:
Taking a look at the `engine_scripts`looks like we are only using [puppeteer](https://github.com/CartoDB/airship/pull/142/files#diff-a6a90f152b3c495498c718250396bf7bR35) so we could get rid of the different configurations.

Even if we let the pupetteer config I see a lot of boilerplate code there... for example the one to load the squirrell image: 😂

examples/containers/containers-legends.html
```diff
@@ -7,7 +7,7 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Bottom bar</title>
   <link rel="stylesheet" href="../../packages/styles/dist/airship.css">
-  <script src="../../packages/components/dist/airship.js"></script>
+  <!-- <script src="../../packages/components/dist/airship.js"></script> -->
```
IagoLast: why is this commented?

examples/containers/containers-sidebar.html
```diff
@@ -7,7 +7,7 @@
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Sidebar</title>
   <link rel="stylesheet" href="../../packages/styles/dist/airship.css">
-  <script src="../../packages/components/dist/airship.js"></script>
+  <!-- <script src="../../packages/components/dist/airship.js"></script> -->
```
IagoLast: [same](https://github.com/CartoDB/airship/pull/142/files#r208273827)

packages/icons/src/icons/alert-fill.svg
```diff
@@ -1 +1 @@
-<svg width="16px" height="16px" xmlns="http://www.w3.org/2000/svg"><path d="M7.02981595,0.785783696 C7.56582792,-0.261638373 8.43457874,-0.262217384 8.97088702,0.785783696 L15.7861356,14.1034746 C16.3221476,15.1508966 15.8151156,16 14.6616325,16 L1.33907042,16 C0.182012546,16 -0.321740933,15.1514757 0.214567348,14.1034746 L7.02981595,0.785783696 Z M7.32190618,12.806511 L7.32190618,13.8710073 L8.36629771,13.8710073 L8.36629771,12.806511 L7.32190618,12.806511 Z M7.32190618,6.41953289 L7.32190618,11.7420146 L8.36629771,11.7420146 L8.36629771,6.41953289 L7.32190618,6.41953289 Z"/></svg>
+<svg width="16px" height="16px" xmlns="http://www.w3.org/2000/svg"><path d="M7.02981595,0.785783696 C7.56582792,-0.261638373 8.43457874,-0.262217384 8.97088702,0.785783696 L15.7861356,14.1034746 C16.3221476,15.1508966 15.8151156,16 14.6616325,16 L1.33907042,16 C0.182012546,16 -0.321740933,15.1514757 0.214567348,14.1034746 L7.02981595,0.785783696 Z M7.32190618,12.806511 L7.32190618,13.8710073 L8.36629771,13.8710073 L8.36629771,12.806511 L7.32190618,12.806511 Z M7.32190618,6.41953289 L7.32190618,11.7420146 L8.36629771,11.7420146 L8.36629771,6.41953289 L7.32190618,6.41953289 Z"/></svg>
```
IagoLast: I think you editor changed this when opened

DEVELOPERS.md
ivanmalagon: 👍

.gitignore
ivanmalagon: :+1:

packages/styles/src/core/containers/test/containers-in-sidebar-bottom-and-legends.html
```diff
@@ -25,7 +25,7 @@
           <div class="as-panel as-panel--top as-panel--right">
             <div class="as-panel__element">
               <section class="as-box">
-                <p class="as-title">LEGEND 1a</p>
+                <p class="as-title">LEGENDICA 1a</p>
```
IagoLast: 👀

DEVELOPERS.md
ivanmalagon: `test:build-references`?? I like the explicit mention of references.

backstop.config.js
ivanmalagon: :+1:

backstop.config.js
ivanmalagon: :+1:

packages/styles/src/test-utils.js
```diff
@@ -1,33 +1,25 @@
...
-    screenshot = options.mobile ? `${this.image(name)}-mobile-out.png` : `${this.image(name)}-out.png`;
-    url = this.html(name);
-    return {
-      reference,
-      screenshot,
-      url,
-      viewportWidth: options.mobile ? mobileWidth : desktopWidth,
-      viewportHeight: options.mobile ? mobileHeight : desktopHeight
+  this.spec = function (label, filename, readySelector) {
+    let options = {
```
IagoLast: I think we could edit this to pass only a `testName` as a parameter, and use `testName` as label and `testName.html` as a url.

This way we only need to pass a single parameter to the tests.

examples/containers/containers-legends.html
ivanmalagon: The loader doesn't behave with static files. They need to be served.

And we don't need the components anyway in `styles` package tests.

packages/styles/src/core/containers/test/containers-in-sidebar-bottom-and-legends.html
ivanmalagon: It's was a trap to check if you were paying attention.

IagoLast:
@rubenmoya @ivanmalagon crazy idea but... can we use netlify to publish the visual UI test results webpage ? so we dont need to run the tests locally again!

examples/containers/containers-legends.html
IagoLast: so its better to remove the line right?

IagoLast:
I'm not sure about the last changes... why do we need a threshold? If we are using the same browser/machine for all tests, why the references are different?

ivanmalagon:
There were inconsistencies because of animation when a radio button and checkbox is checked.

I've reduced the global delay, increased the delay for those tests and adjusting the rest to avoid false negatives.

IagoLast:
🚀 we can do [this](https://github.com/CartoDB/airship/pull/142#discussion_r208248758) in a new issue :)

ivanmalagon:
It's already possible as you can see in https://github.com/CartoDB/airship/pull/142/files#diff-6a3953e8687a89c5a31e57af5a04ca8c

Anyway, I've left open the chance to set a label different than the filename. And a optional object with scenario options.

PR #143, opened 2018-08-06
43 checkbox

IagoLast:
#43 Checkbox
---


packages/styles/src/checkbox/checkbox.scss
```diff
@@ -44,11 +44,16 @@
     margin: 0;
     padding: 0;
     border-radius: 3px;
+    outline: none;
     background: none;
     cursor: pointer;
 
     &:focus {
-      border-width: 2px;
+      border: 2px solid $color-primary;
```
VictorVelarde: @IagoLast Should we use this instead: `border: 2px solid var(--as-color-primary, #{--as-color-primary});`

PR #146, opened 2018-08-08
Add tooltips

IagoLast:
#43 Tooltips
---


piensaenpixel:
<3

packages/styles/src/tooltip/README.md
```diff
@@ -0,0 +1,31 @@
+Tooltips are small, interactive, textual hints for mainly graphical elements.
+
+Notice we only provide the `styles` and `css classes` required to display a tooltip you need to control the placement
```
VictorVelarde: required. To display a tooltip

packages/styles/src/tooltip/tooltip.scss
```diff
@@ -0,0 +1,44 @@
...
+  &::after {
+    content: "";
+    position: absolute;
+    border-width: 5px;
+    border-style: solid;
+    border-radius: 2px;
+    border-color: var(--as-color-type-01, #{$color-type-01});
+  }
+
+  &--bot::after {
```
VictorVelarde: maybe we could use `bottom`instead of `bot`?

IagoLast: we use bot all along the project

VictorVelarde: so bot-it-is 🤖

Issue #148, opened 2018-08-08
UI tests improvements


IagoLast:
- [x] Remove old testing scripts
- [x] Clean `package.json` files
- [ ] Make backstop.js logs cleaner
- [x] Make tests stable waiting for the page to load instead using a delay

cc @ivanmalagon

ivanmalagon:
Can we close this @IagoLast ? What did you mean with `make backstop.js logs cleaner`?

IagoLast:
We can close it.

> What did you mean with make backstop.js logs cleaner?

Meh, I meant backstop logs lots of noise, but its ok

PR #149, opened 2018-08-09
148 ui test improvements

IagoLast:
#148 
---


IagoLast:
LGTM, we still need the delay because even when the fonts loaded there are some animations in the ui-tests that need to be completed before taking an screenshot

PR #152, opened 2018-08-13
211 Dropdown menu

IagoLast:
#211 
---


IagoLast:
I also need a clarification here, is this a `menu` or a `select` ?

ivanmalagon:
Both should be done.

A style to show a contextual menu and a dropdown input.

jesusbotella:
I have been taking a look at the documentation and I think styles are not OK. Text inside the dropdown is not aligned to left (with a left margin), list separator colour is not the same one we have in Sketch, and I am not sure but I think font size is not exactly the same as in Sketch either.

packages/styles/src/menu-dropdown/menu-dropdown.scss
```diff
@@ -0,0 +1,95 @@
+@import "../core/_dev";
+
+.as-menu-dropdown {
+  --as-dropdown-menu-hover-color: #D3F6E6;
```
jesusbotella: Shouldn't we put a SASS variable to set the colour in the CSS variable?

packages/styles/src/menu-dropdown/menu-dropdown.scss
```diff
@@ -0,0 +1,95 @@
...
+    border: 0;
+    background: var(--as-color-ui-01, #{$color-ui-01});
+    color: var(--as-color-primary, #{$color-primary});
+    font: $font-body;
+    text-align: left;
+    cursor: pointer;
+
+    &:hover,
+    &:focus {
+      // Pending focus
```
jesusbotella: Still pending 😂

packages/styles/src/menu-dropdown/menu-dropdown.scss
```diff
@@ -0,0 +1,95 @@
...
+        border-radius: 4px 4px 0 0;
+      }
+
+      &:last-child {
+        border-radius: 0 0 4px 4px;
+      }
+
+      &:hover,
+      &:focus-within {
+        background: var(--as-dropdown-menu-hover-color, #D3F6E6);
```
jesusbotella: SASS variable for fallback?

packages/styles/src/menu-dropdown/menu-dropdown.scss
```diff
@@ -0,0 +1,95 @@
...
+      }
+
+      button {
+        width: 100%;
+        height: 100%;
+        min-height: 32px;
```
jesusbotella: I don't know if we're using spacing units for height in other components, but we should, right?

IagoLast: what do you mean?

jesusbotella: I mean that I don't know if we should put `min-height: get-spacing(8)`.  I don't think if we have ever used that on height/width, but I don't know if we should.

packages/styles/src/menu-dropdown/menu-dropdown.scss
IagoLast: yep, need design opinion about this

packages/styles/src/menu-dropdown/menu-dropdown.scss
IagoLast: since this color is only used here I'm not sure if we should create a sass variable

packages/styles/src/menu-dropdown/menu-dropdown.scss
IagoLast: [same](https://github.com/CartoDB/airship/pull/152#discussion_r214301170)

jesusbotella: You've used it twice already xD
I think we can put it as a SASS variable on the top of the file with that colour.
It is `#47DB99` (which is an Airship color) with an alpha channel of 16, but as far as I know there's not a syntax for `rgba(#47DB99, 16)`. I don't know if we can interpolate a string like `#47DB99FF` with `#{$color-complementary}FF` to a variable and use it as background.

packages/styles/src/menu-dropdown/menu-dropdown.scss
IagoLast: cc @CartoDB/design

IagoLast:
@jesusbotella 👀 plz

PR #153, opened 2018-08-13
Add basic tables

IagoLast:
#43 
---


rubenmoya:
I think the green color in rows is because of `tr:hover` not becuase of `tr:even`

IagoLast:
I think `:hover` is slightly darker

rubenmoya:
There are two `:hover` effects, one for the `tr` and one for the `td`, both of them using opacity to get that effect (the cell being darker than the row), but the green color should be used only in hover effects, not in even rows

IagoLast:
@rubenmoya fixed :)

rubenmoya:
👍

PR #155, opened 2018-08-14
Add switch element

IagoLast:
#43 
---


packages/styles/src/switch/switch.scss
```diff
@@ -0,0 +1,80 @@
+@import "../core/_dev";
+
+.as-switch {
```
jesusbotella: I miss some disabled styles and the disabled example in Catalog. I don't think the current disabled styles match specification.

packages/styles/src/switch/switch.scss
```diff
@@ -0,0 +1,80 @@
...
+  display: inline-block;
+  position: relative;
+
+  svg {
```
jesusbotella: We're not using BEM here, this selector could be `.as-switch--icon`. Not sure about the ones below, but I think we shouldn't be using element selectors as we agreed in other places.

packages/styles/src/switch/switch.scss
```diff
@@ -0,0 +1,80 @@
...
+    position: absolute;
+    top: 4px;
+    left: 4px;
+    transition: opacity 0.2s ease;
+    opacity: 0;
+    color: $color-ui-01;
```
jesusbotella: Aren't we going to use CSS Variables here? More examples of this issue below.

packages/styles/src/switch/switch.scss
```diff
@@ -0,0 +1,80 @@
...
+    padding: 0;
+    overflow: hidden;
+    clip: rect(0 0 0 0);
+    border: 0;
+  }
+
+  input + label {
+    display: block;
+    position: relative;
+    width: 32px;
```
jesusbotella: Spacing measures?

IagoLast:
> What about the input

The input can only be an input :trollface:

packages/styles/src/switch/switch.scss
IagoLast: fixed

rubenmoya:
Hi! Sorry for jumping in late, but the color for the disable border should be `#E2E6E3`, which is `$color-ui-03` in our variables, and the circle should be `#D1D5D7` which is `$color-ui-04` in our variables.

Issue #96, opened 2018-07-25
Modal


ivanmalagon:
![image](https://user-images.githubusercontent.com/1078228/43207402-5074c2da-9028-11e8-8f9d-b6d665c4e0b2.png)

We need to provide CSS for a modal window (scrim + modal container)

### Questions
- What color / opacity should apply to the scrim background? @CartoDB/design

urbanphes:
Color: `#2C2C2C`
Opacity: 80%

Example:
![10_traffico_tiempo-real](https://user-images.githubusercontent.com/14546701/43388745-3b2fe40c-93ea-11e8-892e-4e5b3a66f869.png)

PR #162, opened 2018-08-16
Add modal element

IagoLast:
#96 
---


packages/styles/src/modal/modal.scss
```diff
@@ -0,0 +1,21 @@
...
+  right: 0;
+  bottom: 0;
+  left: 0;
+  align-items: center;
+  justify-content: center;
+  background: rgba(44, 44, 44, 0.8);
+
+  &__content {
+    max-width: 450px;
+    max-height: 80vh;
```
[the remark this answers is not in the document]
IagoLast: invoking @piensaenpixel for a decision

packages/styles/src/modal/modal.scss
```diff
@@ -1,19 +1,23 @@
 @import "../core/_dev";
 
 .as-modal {
+  --as-modal-background: rgba(44, 44, 44, 0.8);
```
rubenmoya: This should be `rgba($as-color-type-02, 0.8)`, same below

packages/styles/src/modal/modal.scss
```diff
@@ -1,19 +1,23 @@
...
+
   display: flex;
   position: absolute;
+  z-index: 2;
```
rubenmoya: Is `2` enough? should we use a higher value?

packages/styles/src/modal/modal.scss
```diff
@@ -1,19 +1,23 @@
...
   left: 0;
   align-items: center;
   justify-content: center;
-  background: rgba(44, 44, 44, 0.8);
+  background: var(--as-modal-background, rgba(44, 44, 44, 0.8));
 
   &__content {
     max-width: 450px;
     max-height: 80vh;
+    padding: get-spacing(4);
```
rubenmoya: In the invision project, the padding for the modals is `40px`

packages/styles/src/modal/test/modal.html
```diff
@@ -28,7 +28,7 @@ <h1 class="as-title">This is a modal</h1>
         natus voluptatem praesentium itaque sunt commodi odio incidunt fugiat, cupiditate ipsum officiis eligendi?
       </p>
       <hr>
-      <h2 class="as-subheader">I have no padding</h2>
+      <h2 class="as-subheader">I have padding</h2>
```
rubenmoya: Should we show a real example instead of a lorem ipsum? Try to copy the invision example or something like that

packages/styles/src/modal/modal.scss
IagoLast: which value is ok?

rubenmoya: I have no idea, but we're going to have the same problem with dropdowns, so we should define a list of z-index(es)

IagoLast:
@rubenmoya ready to 🚀  ?

Issue #78, opened 2018-07-19
Layout containers: Infowindow container


ivanmalagon:
### Mockups

![image](https://user-images.githubusercontent.com/1078228/42955619-33c452ee-8b7e-11e8-84eb-0aa6eddfbd9e.png)

### Characteristics

Width: 260px

Content can be layed out using the CSS built on https://github.com/CartoDB/airship/issues/40

PR #163, opened 2018-08-16
#78 infowindow

IagoLast:
#78 
---

After discussing this with @rubenmoya and @jesusbotella we found this issue is very hard to implement.

We created [a lightweight (simple to develop) alternative](https://deploy-preview-163--carto-airship.netlify.com/catalog/#/styles/infowindow) but we need to know the design opinion first.

Limitations: 

- The hook is different (no idea how to fix this) [0][1]
- Image infowindow has no shadow (we could fix this but we need js, loading 2 images and the markup gets complicated) [1]
- Image infowindow has no border-radius on the bottom-right corner (we could do a fine-tunning in the clip-path to achieve this) [1]

cc @ivanmalagon 

[0]
screen shot 2018-08-16 at 16 36 18

[1]
screen shot 2018-08-16 at 16 35 48


rubenmoya:
IMHO this should be a web component since that will solve all of the corner cases and will "just work"

Issue #161, opened 2018-08-14
Layout manager component


ivanmalagon:
In the documentation for the layout components and it's responsive behaviour we needed to add JS code for showing the tabs, controlling the panels and collapsing the toolbar items.

It would be great if all that was encapsulated in a component that takes care of it, freeing the developer from writing that code for every application and working out of the box.

### Goal
- [ ] Make a web component that manages the layout as we do in the layout examples.

This is kind of research so feel free to discuss and create the API you think it's more appropiate.

jesusbotella:
I created a component to make layout composition with Airship a bit easier, without having to code UI manipulation methods to hide or show different content sections.

The component is `<as-application-content></as-application-content>`.

The way it works is: you put your application content inside the component, and it automatically detects its sections and create tabs accordingly.

For example:
```html
<as-aplication-content>
  <aside class="as-sidebar as-sidebar--left">Left Sidebar</aside>
    
  <div class="as-map-wrapper">
    <div class="as-map">
      <div id="map"></div>
      <div class="as-panels">
        <div class="as-panel as-panel--top as-panel--right">
          <div class="as-panel__element">Floating Panel</div>
        </div>
      </div>
    </div>
    
    <div class="as-bottom-bar">Bottom Bar</div>
  </div>
    
  <aside class="as-sidebar as-sidebar--right">Right Sidebar</aside>
</as-aplication-content>
```
This content will create 5 tabs according to the content. The tabs for the map, both sidebars, floating panels, and the bottom bar. These tabs will be rendered with the same code we had, and will be listening for click events that will change content when occurring.

Each of these tabs can be named and ordered as we want, using data attributes. We can use `data-name` and `data-tab-order` to override default name and ordering like this:

```html
<div class="as-bottom-bar" data-name="Histogram" data-tab-order="1">
  Bottom Bar
</div>
```

There are two public methods:

- **getSections()**
This method return a list of the sections discovered throughout the content with some interesting properties, such as:
```typescript
interface ApplicationSection {
  active?: boolean;
  activeAction?: (section: ApplicationSection) => {};
  activeClass?: string;
  disableAction?: (section: ApplicationSection) => {};
  element: HTMLElement;
  name: string;
  type: string;
  tabOrder: number;
}
```

- **setVisible(sectionName)**
This method changes visible content section by introducing the section name like:

```js
    const applicationContent = document.querySelector('as-application-content');
    aplicationContent.setVisible('Histogram');
```

[...]

PR #164, opened 2018-08-17
Layout Manager

jesusbotella:
Please read #161 for more context. I explained everything there.


ivanmalagon:
Let's use good looking styles, like the ones used in `containers` section. I don't feel the rendered example pass our design standards.

Apart from that, legends are being rendered weirdly. Get rid of the middle legends. With the two at the right top corner is enough.

image

PR #169, opened 2018-08-21
Remove margin in badges

rubenmoya:
The badges are not aligned with the content because of the margin

screen shot 2018-08-21 at 11 20 15


rjimenezda:
Ivan's commit fixes everything, LGTM

Issue #172, opened 2018-08-21
No catalog link in README.md


ivanmalagon:
I couldn't find clearly our netlify preview in the master branch.

Add `https://carto-airship.netlify.com/#/` to README.md

PR #197, opened 2018-08-22
Add link to documentation on the README

rjimenezda:
#172

We're linking on the repo description to the website, where there's a lin to catalog, but this adds it to the README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 [![CircleCI](https://circleci.com/gh/CartoDB/airship/tree/master.svg?style=svg)](https://circleci.com/gh/CartoDB/airship/tree/master)
 
-There are two npm packages for easier use:
+There are three npm packages for easier use:
 
 
 | module | status | version | description |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,6 +12,10 @@ There are two npm packages for easier use:
 |  [airship-components](https://www.npmjs.com/package/@carto/airship-components) | ![status](https://img.shields.io/badge/status-in%20progress-orange.svg) | [![npm version](https://badge.fury.io/js/%40carto%2Fairship-components.svg)](https://badge.fury.io/js/%40carto%2Fairship-components) | Web components for Location Intelligence apps. |
 |  [airship-icons](https://www.npmjs.com/package/@carto/airship-icons) | ![status](https://img.shields.io/badge/status-in%20progress-orange.svg) | [![npm version](https://badge.fury.io/js/%40carto%2Fairship-icons.svg)](https://badge.fury.io/js/%40carto%2Fairship-icons) | High quality icons set |
 
+## Documentation
+
+The library is documented with [Catalog](https://www.catalog.style), check out our [documentation website](https://carto-airship.netlify.com/#/)
```
rubenmoya: Is it necessary to say what we use to build the documentation?

rjimenezda: No, it's just that it felt too short without it.

I'll remove the catalog bit and just leave the Check out our blabla

We should ask ourselves if we want to have some short documentation here as well, as you mentioned.

Issue #178, opened 2018-08-21
npm release should build first


ivanmalagon:
`npm run release` does not run `npm run build`. Add it to the workflow.

PR #198, opened 2018-08-22
Run build before lerna release

rjimenezda:
#178

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "lint:fix": "lerna run lint:fix",
     "lint": "lerna run lint",
     "postinstall": "lerna bootstrap --hoist",
-    "release": "lerna publish --npm-tag prerelease",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "lint:fix": "lerna run lint:fix",
     "lint": "lerna run lint",
     "postinstall": "lerna bootstrap --hoist",
-    "release": "lerna publish --npm-tag prerelease",
+    "release": "npm run build && lerna publish --npm-tag prerelease",
```
rubenmoya: Shouldn't it be something like this?

- Release: Bump the version
- Prepublish hook: Run build command
- Publish: Upload to NPM

rubenmoya: Yep, it seems to be something related to lerna, forget about my comment

Issue #173, opened 2018-08-21
Publish in our own CDN


ivanmalagon:
Airship CSS is used in the examples using jsdelivr. It has an aggressive cache mechanism so a new published version is not available until 24h has passed.

Add our own CDN `https://libs.cartocdn.com/` to our release process for the three packages: components, icons, style.

jimaek:
Hey, jsDelivr offers an easy to use cache purge API to deploy new versions. You can use it during your deployments and completely automate the process

```
CDN URL:
https://cdn.jsdelivr.net/npm/jquery@3.2.1/AUTHORS.txt

Purge URL:
https://purge.jsdelivr.net/npm/jquery@3.2.1/AUTHORS.txt


Multiple files (20 limit):

curl -X POST \
  http://purge.jsdelivr.net/ \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json' \
  -d '{
	"path": [
		"/npm/test",
		"/npm/test2",
		"/npm/test3"
	]
}'
```

PR #204, opened 2018-08-23
S3 upload

rjimenezda:
#173

I've basically made a simpler version of grunt-aws (only for S3) and grunt-fastly

The node script will upload all files on each package's release folder to a corresponding "folder" in S3. We still have to ask infra to make this work for us.

If there's no prerelease on the version, I'm uploading every file three times. For example:

If we have v1.2.3 of airship-components, on the CDN there'll be:

=> airship-components/v1/...
=> airship-components/v1.2/...
=> airship-components/v1.2.3/...

There's a --verbose and --dry options for debugging. So feel free to test them out


scripts/release-s3.js
```diff
@@ -0,0 +1,146 @@
...
+      }
+
+      for (vers of versions) {
+        const key = `${destination}/v${vers}/${dst}`;
+        try {
+          if (!DRY_RUN) {
+            await putObject({
+              Key: key,
+              Bucket: secrets.AWS_S3_BUCKET,
+              Body: fileContent
```
Jesus89: Maybe you want to add here `ACL: 'public-read'`

rjimenezda:
I've added the ACL that @Jesus89 mentioned, Content-Encoding for zipped files, Content-type otherwise they're served as octet-stream, and some more debug info, like the compression ratio. There's only two files that grow a bit when zipped, and I'm skipping WOFF files because they're already zipped.

I've uploaded the prerelease and it seems to work fine.

release.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "Styles v$COMPONENTS_VERSION"
+echo "Are the version numbers bumped?"
+read -n1 -r -p "Press Ctrl+C to cancel, or any other key to continue." key
+
+npm build
+
+echo "Uploading to CDN..."
+
+node scripts/release-s3 --dry --verbose
+nose scripts/invalidate-cache --dry --verbose
```
ivanmalagon: 👃

Issue #205, opened 2018-08-23
Histogram issues


ivanmalagon:
#### Format axis

![image](https://user-images.githubusercontent.com/1078228/44548056-48ad3080-a71d-11e8-84a8-56c305b5cd43.png)

Numbers on axis should be formatted to avoid getting numbers overflown.

#### Back to black

When selecting a range and then clearing the selection, the button gets black.

![back to black](https://user-images.githubusercontent.com/1078228/44663144-fcf5d200-aa0f-11e8-824c-d049be749f37.gif)

rjimenezda:
So I sort of fixed this, but then @alexsroth pointed out that it sometimes doesn't look that great either.

Case in point:

![image](https://user-images.githubusercontent.com/446970/44595435-34bf0880-a7c9-11e8-9497-0c36c898760e.png)

We should probably play with d3-axis which in turn uses d3-format, to get an acceptable number format that's not too long.

jesusbotella:
Regarding back to black, we set the `:focus` to the same color as the active one so that was the issue. If we want to differentiate the focus state, I'd put an outline like browsers do.

![histogrambutton](https://user-images.githubusercontent.com/4319728/44667756-f40afd80-aa1b-11e8-8ff1-bcd21338e585.gif)

jesusbotella:
Fixed!

PR #208, opened 2018-08-27
Fix some histogram issues

rjimenezda:
#205
---

I fixed two problems, the selector overflowed a few pixels on the right, so I made it visible, and I've added d3-format as an explicit dependency to fine tune the axis format.

image

The `0k` was driving me insane :upside_down_face:, it seems like d3-axis has some issues when specifying a string format instead of the d3-format function.


rjimenezda:
While writing an example with Vue I've noticed a couple of problems, so I adjusted the SVG margin and made the fonts consistent with the category widget

PR #243, opened 2018-09-04
233 guides getting started

### LINKED ISSUE #233: Guides: Introduction
Create an introduction guide with the following contents:

- What is
- What for
- Who
- Expectations

IagoLast:
closes #233 
---


doc/guides/01-getting-started.md
```diff
@@ -0,0 +1,178 @@
...
+
+    npm i @carto/airship-style @carto/airship-icons
+
+Take a look at [sass-loader](https://github.com/webpack-contrib/sass-loader) if you are using webpack. We provide both `dist` and `src` files so you can create a custom [sass](https://sass-lang.com/) build.
+
+### Web components
+
+    npm i @carto/airship-components
+
+You need to call the `defineCustomElements` function to use the web components from npm.
```
jesusbotella: I'd suggest to execute `defineCustomElements` in the main script to avoid issues :)

doc/guides/01-getting-started.md
```diff
@@ -0,0 +1,178 @@
...
+  <link rel="stylesheet" href="https://libs.cartocdn.com/airship-styles/dist/utilities/utilities.css">
+<head>
+```
+
+
+## Usage from NPM
+
+### Styles and fonts
+
+Just install our packages and use it as you wish:
```
jesusbotella: We are not telling people how to import the whole style/icons bundle or separate styles in their application

doc/guides/01-getting-started.md
```diff
@@ -0,0 +1,178 @@
...
+
+defineCustomElements(window);
+```
+
+# Using Airship
+Using airship once included in your webpage is very straightforward. 
+
+## Airship Styles
+
+To use airship styles you only need to add CSS classes to the desired HTML fragment.
```
jesusbotella: `To use airship styles**,** you only need to add CSS classes to your HTML`

doc/guides/01-getting-started.md
```diff
@@ -0,0 +1,178 @@
...
+
+For example a button can become a Airhip button adding the `as-btn` class.
```
jesusbotella: Airhip

A simple button can become a Airship button by simply adding `as-btn` class

doc/guides/01-getting-started.md
```diff
@@ -0,0 +1,178 @@
...
+<button class="as-btn"> Button </button>
+```
+
+Airship classes follow [BEM](http://getbem.com/introduction/) conventions, this means that the classes can be of the following types:
+
+- **Block**: Standalone entity that is meaningful on its own. (`.block`)
+- **Element**: A part of a block that has no standalone meaning and is semantically tied to its block. (`.block__element`)
+- **Modifier**: Modifies the block appearance or behavior. (`.block--modifier`)
+
+For example to create tabs we use the following classes:
```
jesusbotella: Too many for examples

doc/guides/01-getting-started.md
```diff
@@ -0,0 +1,178 @@
...
+    <button role="tab" class="as-tabs__item">PANELS</button>
+  </div>
+```
+
+## Airship Icons
+You can use airship icons in two different ways.
+
+### As web font
+
+To display an icon only a class and an `i` tag is required.
```
jesusbotella: I'd just put the example here

doc/guides/01-getting-started.md
```diff
@@ -0,0 +1,178 @@
...
+
+```html
+<i class="as-icon-twitter"></i>
+```
+
+
+### As svg
```
jesusbotella: SVG please :)

doc/guides/01-getting-started.md
```diff
@@ -0,0 +1,178 @@
...
+
+```html
+<img src="https://libs.cartocdn.com/airship-icons/<VERSION>/icons/twitter.svg" alt="Twitter logo">
+```
+
+## Web componenets
+
+Web components should be treated as regular HTML Elements. 
+
+A web component is just an HTML tag with some attributes that control it's behaviour. Simply include the HTML tag, and edit its properties through attributes or through javascript.
```
jesusbotella: `its`

doc/guides/01-getting-started.md
IagoLast: I dont follow you here...

doc/guides/01-getting-started.md
IagoLast: what do you mean?

doc/guides/01-getting-started.md
IagoLast: any suggestion?

doc/guides/01-getting-started.md
jesusbotella: What I meant is that the users can import the components in whichever part of the application that they want, but I would suggest the users to import them in the main "entry point" to avoid possible issues.

doc/guides/01-getting-started.md
jesusbotella: I mean that we do not tell the users how to import the whole styles or separate styles when installing it from npm as we do with the CDN in https://github.com/CartoDB/airship/blame/f20fa90a19f65a6af9d9d16d6474f4ee459798b3/doc/guides/01-getting-started.md#L14-L44

doc/guides/01-getting-started.md
jesusbotella: I mean that the text is redundant with the example below, or I would change the wording at least.

doc/guides/01-getting-started.md
IagoLast: well, actually next line suggest you can create a custom build:

> Take a look at [sass-loader](https://github.com/webpack-contrib/sass-loader) if you are using webpack. We provide both `dist` and `src` files so you can create a custom [sass](https://sass-lang.com/) build.

jesusbotella: That's a vague description. I think we should tell the users, at least, that they can import styles with
```js
import '@carto/airship-styles';
```

Use it as you wish is not a concrete way of using Airship, we should suggest one at least.

PR #256, opened 2018-09-06
Add web components guide

### LINKED ISSUE #237: Guides: Web components
Create a guide explaining the basics of Airship web-components

- Overview
- What is a web component
- How to use a web component
- List of web components
- Browser compatibility
- Integration with another framework

IagoLast:
Fix #237 
---

### DIFF
--- a/docs/guides/03-web-components.md
+++ b/docs/guides/03-web-components.md
@@ -1,0 +1,94 @@
+## Web components
+
+Airship provides a set of [web components](https://www.webcomponents.org/introduction). One of the key features of the Web Components standard is the ability to offer custom elements that encapsulate functionality on an HTML page, rather than having to make do with a long, nested batch of elements that together provide a custom page feature.
+
+This is especially interesting to create geospatial widgets, such as a histogram or a category-view where instead of writing all the code by hand every time you can use just a simple a custom element.
+
+
+```html
+<as-category-widget
+  heading="Business Volume"
+  description="Description"
+  default-bar-color="#47DB99">
+</as-category-widget>
+```
+
+### Attributes
+
+[Attributes](https://en.wikipedia.org/wiki/HTML_attribute) modify the default functionality of an element. Usually the attributes will have a default value and using them will be optional. In the example above, heading and description are required attributes while the bar color is optional. 
+
+HTML Attributes can be only of type `string`, so if you want to use an object as an attribute you need to use javascript.
+
+Although you can use the [setAttribute](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute) method, you can directly assign the attributes as properties of the element.
+
+```js
+const $categoryWidget = document.querySelector('as-category-widget');
+const categoriesList = [
+  { name: 'A New Hope', value: 359029623, color: '#FABADA' },
+  { name: 'The Empire Strikes Back', value: 236513856 },
+  { name: 'Return of the Jedi', value: 204338075 },
+];
+
+// Same as: $categoryWidget.setAttribute('categories', categoriesList);
+$categoryWidget.categories = categoriesList;
+```
+
+### Loading web components
+
+Loading and using web components is very easy, you just need to include our component loader in the head of your app. Once the loader was included It will automatically detect the use of web-components and load them on demand. It also detects browser capabilities and will automatically decide whether or not to include polyfills, leading to smaller bundles in modern browsers.
+
+```html
+<head>
+  <!-- Include the loader on the header -->
+  <script src="https://libs.cartocdn.com/airship-components/<VERSION>/airship.js"></script>
+</head>
+<body>
+  <!-- Use Web components as a regular HTML Elements and the loader will take care of the rest -->
+  <as-histogram-widget></as-histogram-widget>
+</body>
+```
+
+> The loader may conflict with the webpack bundler if the [publicPath](https://webpack.js.org/guides/public-path/) property is not configured correctly.
+
+### List of Web Components
+
+> Link to the reference
+
+- Widgets
+  - category-widget
+  - histogram-widget
+- Components
+  - infowindow
+  - range-slider
+- Layout
+  - application-content 
+  - toolbar
+  - toolbar-action
+
+### Framework integration
+
+You can use Airship components with another web frameworks like Angular, React or Vue. See [https://stenciljs.com/docs/framework-integration](https://stenciljs.com/docs/framework-integration).
+
+
+
+### Browser support
+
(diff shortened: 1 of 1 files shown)


docs/guides/03-web-components.md
```diff
@@ -0,0 +1,94 @@
+## Web components
+
+Airship provides a set of [web components](https://www.webcomponents.org/introduction). One of the key features of the Web Components standard is the ability to offer custom elements that encapsulate functionality on an HTML page, rather than having to make do with a long, nested batch of elements that together provide a custom page feature.
+
+This is especially interesting to create geospatial widgets, such as a histogram or a category-view where instead of writing all the code by hand every time you can use just a simple a custom element.
```
jesusbotella: Did you mean "category widget"?

jesusbotella: I'd say "from scratch" instead of "by hand"

docs/guides/03-web-components.md
```diff
@@ -0,0 +1,94 @@
...
+<as-category-widget
+  heading="Business Volume"
+  description="Description"
+  default-bar-color="#47DB99">
+</as-category-widget>
+```
+
+### Attributes
+
+[Attributes](https://en.wikipedia.org/wiki/HTML_attribute) modify the default functionality of an element. Usually the attributes will have a default value and using them will be optional. In the example above, heading and description are required attributes while the bar color is optional.
```
jesusbotella: "Usually" needs to be after the auxiliary verb

docs/guides/03-web-components.md
```diff
@@ -0,0 +1,94 @@
...
+  description="Description"
+  default-bar-color="#47DB99">
+</as-category-widget>
+```
+
+### Attributes
+
+[Attributes](https://en.wikipedia.org/wiki/HTML_attribute) modify the default functionality of an element. Usually the attributes will have a default value and using them will be optional. In the example above, heading and description are required attributes while the bar color is optional. 
+
+HTML Attributes can be only of type `string`, so if you want to use an object as an attribute you need to use javascript.
```
jesusbotella: HTML Attributes' value can **only** be strings.

What do you think?

docs/guides/03-web-components.md
```diff
@@ -0,0 +1,94 @@
...
+
+Although you can use the [setAttribute](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute) method, you can directly assign the attributes as properties of the element.
```
jesusbotella: You can assign the attributes to the element properties, or use [setAttribute](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute) method to do it in a different way.

Do you think it's better?

docs/guides/03-web-components.md
```diff
@@ -0,0 +1,94 @@
...
+  { name: 'Return of the Jedi', value: 204338075 },
+];
+
+// Same as: $categoryWidget.setAttribute('categories', categoriesList);
+$categoryWidget.categories = categoriesList;
+```
+
+### Loading web components
+
+Loading and using web components is very easy, you just need to include our component loader in the head of your app. Once the loader was included It will automatically detect the use of web-components and load them on demand. It also detects browser capabilities and will automatically decide whether or not to include polyfills, leading to smaller bundles in modern browsers.
```
jesusbotella: Once the loader is included, it will automatically detect the use of web-components and load them on demand.

docs/guides/03-web-components.md
```diff
@@ -0,0 +1,94 @@
...
+</head>
+<body>
+  <!-- Use Web components as a regular HTML Elements and the loader will take care of the rest -->
+  <as-histogram-widget></as-histogram-widget>
+</body>
+```
+
+> The loader may conflict with the webpack bundler if the [publicPath](https://webpack.js.org/guides/public-path/) property is not configured correctly.
+
+### List of Web Components
```
jesusbotella: I wouldn't put a list of the components here, we can link to the reference page in DC, which should have one :)

docs/guides/03-web-components.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Framework integration
+
+You can use Airship components with another web frameworks like Angular, React or Vue. See [https://stenciljs.com/docs/framework-integration](https://stenciljs.com/docs/framework-integration).
+
+
+
+### Browser support
+
+Airship Components are build on Stencil, that run natively or near-natively in all widely used browsers.
```
jesusbotella: `are built with`

docs/guides/03-web-components.md
```diff
@@ -0,0 +1,94 @@
...
+
+Web Components are being used in production in 2017 with the above approach.
+
+
+| [<img src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/edge/edge_48x48.png" alt="IE / Edge" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>IE / Edge | [<img src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/firefox/firefox_48x48.png" alt="Firefox" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>Firefox | [<img src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/chrome/chrome_48x48.png" alt="Chrome" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>Chrome | [<img src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/safari/safari_48x48.png" alt="Safari" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>Safari | [<img src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/safari-ios/safari-ios_48x48.png" alt="iOS Safari" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>iOS Safari |
```
jesusbotella: I wouldn't put that *, just for simplicity

docs/guides/03-web-components.md
IagoLast: done

docs/guides/03-web-components.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Framework integration
+
+You can use Airship components with another web frameworks like Angular, React or Vue. See [https://stenciljs.com/docs/framework-integration](https://stenciljs.com/docs/framework-integration).
+
+
+
+### Browser support
+
+Airship Components are build with Stencil, that run natively or near-natively in all widely used browsers.
```
VictorVelarde: @IagoLast 👀 typo --> "built with"

Issue #238, opened 2018-09-04
Guides: CSS Elements


IagoLast:
Create a basic guide with the following contents:

- CSS framework overview
- Basic usage and concepts
- Framework integration
- Browser compatibility
- Basic Customization (We will have an advanced guide talking about airship customization)
  - css variables
  - sass variables
  - inline code

ivanmalagon:
Merged.

PR #260, opened 2018-09-06
CSS elements guide

jesusbotella:
#238

### DIFF
--- a/docs/guides/01-css-elements.md
+++ b/docs/guides/01-css-elements.md
@@ -1,0 +1,127 @@
+# CSS Elements
+
+## Installation
+
+### From NPM
+  #### Install @carto/airship-style package from npm
+  ```
+  npm i @carto/airship-style --save
+  ```
+
+  #### Importing styles in SASS, PostCSS
+  ```css
+  // Import whole bundle
+  @import '~@carto/airship-style';
+
+  // Import SASS element styles
+  @import '~@carto/airship-style/src/{path}';
+
+  // Import CSS element styles
+  @import '~@carto/airship-style/dist/{path}';
+  ```
+
+  #### Importing styles in JavaScript
+  ```javascript
+  // Import whole bundle
+  import '@carto/airship-style';
+
+  // Import SASS element styles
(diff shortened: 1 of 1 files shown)


docs/guides/01-css-elements.md
```diff
@@ -0,0 +1,127 @@
+# CSS Elements
+
```
IagoLast: A small intro giving context will be nice

docs/guides/01-css-elements.md
```diff
@@ -0,0 +1,127 @@
...
+  @import '~@carto/airship-style';
+
+  // Import SASS element styles
+  @import '~@carto/airship-style/src/{path}';
+
+  // Import CSS element styles
+  @import '~@carto/airship-style/dist/{path}';
+  ```
+
+  #### Importing styles in JavaScript
```
IagoLast: should we mention that a specific loader is required?

docs/guides/01-css-elements.md
```diff
@@ -0,0 +1,127 @@
...
+  I'm a regular button
+</button>
+```
+
+Please check [styles documentation]() (missing link) out and see how to use Airship components.
+
+## Framework integration
+To integrate Airship into your framework you only need to import the styles in the way you prefer from the ones stated above.
+
+## Browser Compatibility
```
IagoLast: Usually this section is at the end of the doc

docs/guides/01-css-elements.md
jesusbotella: I don't know, I'd say that we can assume the reader knows how to do that if they use Webpack.

docs/guides/01-css-elements.md
jesusbotella: We can put it at the end of the doc!

docs/guides/01-css-elements.md
jesusbotella: I added a warning just in case any users feel lost about it.

IagoLast:
LGTM

Issue #236, opened 2018-09-04
Guides: Layout


IagoLast:
Create a guide explaining the basics of the airship layout

- Overview
- Grid system
- Responsive breakpoints
- Examples
- Link to templates

jesusbotella:
We need to **clearly** state that if you use `<as-application-container>` you need to put Mapbox.GL/Leaflet code inside `load` event. Well, not only Mapbox.GL/Leaflet but all layout-dependant code.

PR #264, opened 2018-09-10
Update sidebar docs

IagoLast:
#236  Update sidebar reference
---

### DIFF
--- a/packages/styles/src/core/layout/sidebar/README.md
+++ b/packages/styles/src/core/layout/sidebar/README.md
@@ -1,19 +1,109 @@
-## Sidebar
+## .as-sidebar
 
-Sidebars are the area where your app content lives. Sidebars are located beside the map wrapper and should be placed inside an element with the class `as-map-wrapper`. There are no restrictions regarding the contents of a sidebar.
+Use this class to create sidebars beside the map wrapper. Sidebars are the area where your app content lives and should be placed inside an element with the class `as-app-content`. There are no restrictions regarding the contents of a sidebar.
+
+Sidebars are hidden by default in mobile devices and only will be shown when the modifier class `as-sidebar--visible` is present.
+
+
+## Class modifiers
 
 Sidebar position and size can be changed with the following class modifiers.
 
-- Position:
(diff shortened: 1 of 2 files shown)


packages/styles/src/core/layout/sidebar/README.md
```diff
@@ -1,103 +1,137 @@
...
 
 ```html
-noSource: true
-responsive: true
----
-<iframe src="/examples/layouts/sidebar/sidebar.html" style="width: 100%; height: 100%;">
+<div class="as-app"  style="height: 300px; width:100%;">
+  <main class="as-app-content">
+
+    <aside class="as-sidebar as-sidebar--left" style="background: #80B622;"></aside>
```
jesusbotella: Why don't we use utility classes to set all these background color instead of an inline style?

packages/styles/src/core/layout/sidebar/README.md
```diff
@@ -1,103 +1,137 @@
...
-</body>
+</div>
 ```
 
-## Mobile Sidebar
-In order to make a sidebar visible on a mobile phone or any device with a small screen you need to add the `--visible` class modifier you can toggle this class from wherever you want but we recommend using **toolbar-tabs**.
+
+### .as-sidebar--visible
+
+Makes the sidebar visible on small screens occupying all the space available in the`.app-content` and covering the map.
```
jesusbotella: Makes the sidebar visible on small screens filling all the space available in the`.app-content` and covering the map.

#OccupyWallStreet

packages/styles/src/core/layout/sidebar/README.md
```diff
@@ -1,103 +1,137 @@
...
 ---
-<iframe src="/examples/layouts/sidebar/sidebar-nav-bot.html" style="width: 100%; height: 100%;">
+<iframe src="/examples/layouts/sidebar/sidebar.html" style="width: 100%; height: 100%;">
 ```
 
-
 ```code
 lang: html
 ---
+<!-- Missing scripts to handle visibility classes -->
```
jesusbotella: Are we going to include them at some point?

PR #265, opened 2018-09-10
Update map-wrapper docs

IagoLast:
#236 Update map wrapper guide
---


packages/styles/src/core/layout/map-wrapper/README.md
```diff
@@ -1,119 +1,193 @@
...
+</div>
 ```
 
-### Map wrapper + Floating panels
-
-To add floating panels to the map area, to use them for legends for example, you need to add first an element with class `as-panels` inside the `as-map` element.
+## .as-panels
 
-Then you need to add as much elements you want to contains the actual floating panels. The possible positions are set with a combination of place modifiers. For instance, if we want a container to place legends at the top right corner, you need to write:
+Use this class as a child of the `as-map-wrapper` element. The `as-panels` wrapper covers the map area but wont interfer the UI. This element should only be used as a container for the panels.
```
jesusbotella: I think it should be `won't interfere with`

packages/styles/src/core/layout/map-wrapper/README.md
```diff
@@ -1,119 +1,193 @@
...
+You can combine two of the following class modifiers to specify the panel position:
+
 - `as-panel--top`
 - `as-panel--middle`
 - `as-panel--bottom`
 - `as-panel--left`
 - `as-panel--center`
 - `as-panel--right`
 
-You can add an aditional modifier class to allow the panels grow vertically. This way, Airship will stack `as-panel__element` one above the other.
```
jesusbotella: Did we remove `.as-panel--vertical`?

packages/styles/src/core/layout/map-wrapper/README.md
```diff
@@ -1,119 +1,193 @@
...
+
+
+### .as-panel--vertical
+
+By default panels grow horizontally. Add this class to make the panels grow vertically.
+
+```html
+<div class="as-app" style="width:100%; height:300px;">
+  <main class="as-app-content">
+    <div class="as-map-wrapper" style="background: #B5E0F9;">
```
jesusbotella: Can we use an utility class here? ✨

packages/styles/src/core/layout/map-wrapper/README.md
```diff
@@ -1,119 +1,193 @@
-## Map wrapper
+The map wrapper is used as the container for the `map-wrapper` and the `bottom-bar`
```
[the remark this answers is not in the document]
IagoLast: I did this because both developer center and catalog add the header/title at the beginning of the page

packages/styles/src/core/layout/map-wrapper/README.md
IagoLast: nope, but it wasn't documented, good catch :D

IagoLast: nope, is documented below

packages/styles/src/core/layout/map-wrapper/README.md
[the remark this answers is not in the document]
jesusbotella: 👌

Issue #241, opened 2018-09-04
Advanced Guides: Integration with CARTO.js


IagoLast:
Create a guide showing how to integrate airship with CARTO.JS

- Reacting to input changes in the map
- Fetching data from the map

jesusbotella:
Merged!

PR #266, opened 2018-09-10
Advanced CARTO.js guide using Airship

jesusbotella:
#241

### DIFF
--- a/docs/guides/06-advanced-guide-cartojs.md
+++ b/docs/guides/06-advanced-guide-cartojs.md
@@ -196,11 +196,11 @@ Using that Bounding Box filter, our widget will be automatically updated wheneve
       <meta http-equiv="X-UA-Compatible" content="ie=edge">
 
       <!-- Include CSS  -->
-      <link rel="stylesheet" href="https://libs.cartocdn.com/airship-styles/v1.0.0-alpha.39/airship.css">
+      <link rel="stylesheet" href="https://libs.cartocdn.com/airship-styles/v1.0.0-alpha.40/airship.css">
       <!-- Include Icons -->
-      <link rel="stylesheet" href="https://libs.cartocdn.com/airship-icons/v1.0.0-alpha.39/icons.css">
+      <link rel="stylesheet" href="https://libs.cartocdn.com/airship-icons/v1.0.0-alpha.40/icons.css">
       <!-- Include Web Components -->
-      <script src="https://libs.cartocdn.com/airship-components/v1.0.0-alpha.39/airship.js"></script>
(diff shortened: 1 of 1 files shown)


docs/guides/06-advanced-guide-cartojs.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+## Setting a basic layout
+A basic layout might include a container for our map and a sidebar for our future widget.
+
+The layout will always be wrapped by a block element with `.as-app-content` class, which will hold all of our content and elements.
+So, to create our layout we need a right sidebar and a block element wrapped by `.as-map-wrapper` class for our map.
+
+```html
+<section class="as-app-content">
```
IagoLast: `.as-app-content` is normally a `main` tag

docs/guides/06-advanced-guide-cartojs.md
```diff
@@ -0,0 +1,270 @@
...
+
+<!DOCTYPE html>
+<html>
+  <head>
+      <meta charset="UTF-8">
+      <meta name="viewport" content="width=device-width,initial-scale=1.0">
+      <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  </head>
+
+  <body>
```
IagoLast: body should have `as-app` class

docs/guides/06-advanced-guide-cartojs.md
```diff
@@ -0,0 +1,268 @@
...
+
+client.addLayer(layer);
+client.getLeafletLayer().addTo(map);
+```
+
+## Widget creation
+Widgets are the best way to show data from our visualization. You can create a dashboard with a map and several widgets that will allow you to visualize and explore your data in a clearer and better way.
+
+### Adding a Category Widget
+Adding a Category Widget is as simple as including `<as-category-widget>` tag within our `<body>` tag, setting the options you prefer.
```
IagoLast: to the body or to the sidebar?

docs/guides/06-advanced-guide-cartojs.md
```diff
@@ -0,0 +1,268 @@
...
+      <script src="https://libs.cartocdn.com/carto.js/v4/carto.min.js"></script>
+
+      <style>
+        body {
+          margin: 0;
+          padding: 0;
+        }
+
+        .widgets {
+          padding: 10px;
```
IagoLast: I think we should have a default padding

ivanmalagon: Why can't we use a utility class instead of adding a new class here?

jesusbotella:
I have just updated this PR with the Bounding Box filter :)
The relative data in the widget should be ready whenever we merge the fix :)

PR #278, opened 2018-09-13
259 rename cdn

### LINKED ISSUE #259: Standardize package names in npm and CDN
We are using different package names in npm and CDN for styles package.

Right now, it is `@carto/airship-style` in npm and `airship-styles` in CDN.

I suspect that it will be easier to change in CDN than npm but let's see.

IagoLast:
Fixes #259 
---


jesusbotella:
Do we need to remove old scripts?

PR #280, opened 2018-09-14
Fix variable typos

### LINKED ISSUE #279: Typos in _toolbar.scss
There's two colors that are mistyped: 

- `--as-color--secondary`instead of `--as-color-secondary`
- `--as-color--type-04` instead of `--as-color-type-04`

It can cause some issues when overwriting css variables, since you're putting a default value.

IagoLast:
Fix #279 
---


IagoLast:
@rjimenezda you are the reviewer :trollface:

PR #307, opened 2018-09-18
Allow panels to scroll when exceeding max-height

### LINKED ISSUE #292: Panels are not scrollable
Our panels-wrapper covers the full map, so to preserve the interactivity we got the [pointer events](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) disabled.

This prevents the panels to have any kind of scroll.

jesusbotella:
Fix #292


IagoLast:
Failing tests

jesusbotella:
✨

PR #308, opened 2018-09-18
298 recover switch component

### LINKED ISSUE #298: Recover switch components
Our current approach is not intuitive, a component will be much better

IagoLast:
Fix #298 
---


packages/components/src/components/as-switch/README.md
```diff
@@ -0,0 +1,55 @@
+# as-switch
+
+It displays a control to select a range of two numbers from a predefined interval, dragging a slider. Selection can be determined using one or two values. When using one value, the range goes from the minimum to the selected value. When using two, the minimum value can be also configured.
```
jesusbotella: 🤔

packages/components/src/components/as-switch/README.md
```diff
@@ -0,0 +1,55 @@
...
+### Disabled
+```html
+<as-switch disabled id="switch-2" name="switch-2" title="switch-2"></as-switch>
+```
+
+
+## Props
+
+### **checked**: boolean
+The presence of this boolean attribute controls when the switch is checked by default.
```
jesusbotella: The presence of this boolean attribute controls whether the switch is checked by default or not.

packages/components/src/components/as-switch/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+### **disabled**: boolean
+The presence of this boolean attribute controls when the switch is enabled.)
+
+### **name**: boolean
+The name attribute specifies a name for the element. This name attribute can be used to reference the element in a JavaScript.
+
+
+## Events
+All events return an array with one value (simple mode) or two (range).
```
jesusbotella: Please give a review to the whole documentation 😂

packages/components/src/components/as-switch/as-switch.scss
```diff
@@ -0,0 +1,80 @@
+@import '../../../../styles/src/core/_dev';
+
+as-switch {
+  display: inline-block;
```
jesusbotella: Are these styles copied from styles folder? Why don't we reference them to avoid changing the code twice?

packages/components/src/components/as-switch/as-switch.tsx
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * The input name
+   *
+   * @type{string}
+   * @memberof Switch
+   */
+  @Prop() public name: string;
+
+  /**
+   * The input name
```
jesusbotella: I think this is not the input name xD

packages/components/src/components/as-switch/as-switch.tsx
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * @type{string}
+   * @memberof Switch
+   */
+  @Event() public change: EventEmitter;
+
+  /**
+   * When the component is attached to the DOM bind the onCLick function
```
jesusbotella: `onClick` please 🙏

packages/components/src/components/as-switch/as-switch.tsx
```diff
@@ -0,0 +1,88 @@
...
+  public componentDidLoad() {
+    this.el.addEventListener('click', this._onClick.bind(this));
+  }
+
+  /**
+   * Fire a 'change' event with a boolean parameter if the user clicks on an enabled component
+   */
+  public _onClick() {
+    if (!this.disabled) {
+      this.el.querySelector('input').checked = !this.el.querySelector('input').checked;
```
jesusbotella: Let's not query DOM three times.
I think we should update checked prop as well, just in case anyone wants to take it from there.

packages/components/src/components/as-switch/as-switch.tsx
```diff
@@ -0,0 +1,88 @@
...
+      this.change.emit(this.el.querySelector('input').checked);
+    }
+  }
+
+  public render() {
+    return [
+      (<input
```
jesusbotella: I know I told you to put parenthesis when JSX is longer than one line, but they're only necessary when returning JSX not in an array.

packages/components/src/components/as-switch/as-switch.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    it('should emit an event with current status when component toggles', async () => {
+      const spy = _spyOnElement(element);
+
+      element.click();
+
+      expect(spy).toHaveBeenCalled();
+      expect(spy.mock.calls[0][0].detail).toBe(true);
+    });
+
+    it('should emit an event with current status when component toggles', async () => {
```
jesusbotella: This test is named same as the upper one.

I think you can initialize the test with the input checked and then check that it is unchecked correctly.

Actually, it is already checked in a test below those ones.

packages/components/src/components/as-switch/as-switch.scss
IagoLast: They are different :(

IagoLast: anyway we should remove the switch element

jesusbotella: I do think so too! Let's do it.

PR #316, opened 2018-09-21
Fix Range Slider bugs

### LINKED ISSUE #312: Range slider jumps unexpectedly when moving thumb
When using range slider inside a "complex" layout with a left toolbar, it jumps to another position when starting to move the thumb.

airship-slider-bug

### LINKED ISSUE #315: Range slider does not allow decimals
### Context

Setting a slider for opacity like this:

```
            <as-range-slider
              id="opacity-slider"
              min-value="0"
              max-value="1"
              step="0.1"
              value="0.5"></as-range-slider>
```

### Result

opacity

### Expected result
Values are rounded depending on the min and max value. In this case, we should see `0.1 0.2 0.3 ...`

jesusbotella:
This PR fix #315 and fix #312.

Range slider thumb position was being calculated with rail's offsetLeft, and it was not calculated properly when any of the parent elements had any `position: relative`, like `.as-main`. So, this PR uses left offset from `.getBoundingClientRect` which is relative to `document` origin.

Apart from that, we didn't allow decimal numbers in slider's step, and now we do! 🎉

### DIFF
--- a/packages/components/src/components/as-range-slider/slider/as-range-slider.tsx
+++ b/packages/components/src/components/as-range-slider/slider/as-range-slider.tsx
@@ -1,4 +1,5 @@
 import { Component, Event, EventEmitter, Prop, State, Watch } from '@stencil/core';
+import getDecimalPlaces from '../../../utils/get-decimal-places';
 import { Thumb } from '../thumb/as-range-slider-thumb';
 
 @Component({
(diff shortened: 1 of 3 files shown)


packages/components/src/components/as-range-slider/slider/as-range-slider.tsx
```diff
@@ -302,6 +302,27 @@ export class RangeSlider {
...
+    const stepValue = (value / this.step) * this.step;
+    return this.roundToStep(stepValue, this.step);
   }
+
+  private roundToStep(numberToRound: number, step: number) {
+    return Number.parseFloat(numberToRound.toFixed(getDecimalPlaces(step)));
+  }
+}
+
+function getDecimalPlaces(decimalNumber): number {
```
IagoLast: why outside the class?

jesusbotella: I thought It didn't belong to that specific class, but I should've put it in utilities! 😥

PR #317, opened 2018-09-21
Add formatValue function to Category Widget

### LINKED ISSUE #285: Add format to category widget
Allow adding a format function to show, for example, the units to the numeric values in the widget.

jesusbotella:
Fixes #285!

### DIFF
--- a/packages/components/src/components/as-category-widget/as-category-widget.tsx
+++ b/packages/components/src/components/as-category-widget/as-category-widget.tsx
@@ -52,6 +52,14 @@ export class CategoryWidget {
   @Prop() public disableInteractivity: boolean = false;
 
   /**
+   * If this property receives a function, it will be used to format the numbers (eg. for adding $ or €).
+   *
+   * @type {function (value: number)}
+   * @memberof RangeSlider
+   */
+  @Prop() public formatValue: (value: number) => void;
+
+  /**
    * Heading text of the widget
    *
    * @type {string}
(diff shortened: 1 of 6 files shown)


packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -51,6 +51,14 @@ export class CategoryWidget {
    */
   @Prop() public disableInteractivity: boolean = false;
 
+  /**
+   * If this property receives a function, it will be used to format the numbers (eg. for adding $ or €).
```
IagoLast: Function used to format the number output. (e.g. adding currency 10€ or 10$)

packages/components/src/components/as-category-widget/as-category-widget.tsx
```diff
@@ -204,11 +212,13 @@ export class CategoryWidget {
       'as-category-widget__category--selected': isSelected
     };
 
+    const displayValue = this.formatValue ? this.formatValue(category.value) : readableNumber(category.value);
```
IagoLast: This can be missleading, should we pass `readableNumber` to formatValue ? 

- An user see `10` dy default 
- The user creates a format function `number => ${number}€`
- The user will see: `10.00000000€`

jesusbotella: I thought about it the other day and I couldn't come up with the right thing. Some developers would want to have the whole unparsed number and some other might want to have the readable number.

What about returning both values in an object?

packages/components/src/components/as-category-widget/as-category-widget.tsx
jesusbotella: 👌

PR #327, opened 2018-09-24
Add first react guide

### LINKED ISSUE #324: Guides: Integration with react
Create a small guide indicating how to integrate Airship with React

IagoLast:
Fixes #324  
---


docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,111 @@
+We are working on react bindings that offer a better user experience, meanwhile this guide contains the basic fundamentals of how to integrate Airhship components with [React.js](https://reactjs.org/).
```
ivanmalagon: I'd present the two topics in the opposite way:

`This guide contains blah blah blah` and after that sentence `We are working on providing React bindings to offer a better developer experience in the future`

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Web components
+
+A web component is just an HTML tag with some attributes that control its behaviour. Simply include the HTML tag, and edit its properties through attributes or through javascript.
```
ivanmalagon: ... as you would do with a normal HTML element.

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,111 @@
...
+To render a web component simply include its tag inside the REACT render function:
+
+
+```jsx
+render() {
+  <as-infowindow src="https://thecatsite.com/styles/thecatsite/xenforo/sources/notice3.png"></as-infowindow>
+}
+```
+
+The problems come when it comes to managing the state, either listening to events or synchronizing the attributes.
```
ivanmalagon: Don't state it as a problem.

`The challenge with this integration is to manage the state...`
or
`The integration work is to manage the state...`

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,111 @@
...
+    this.ref = React.createRef();
+  }
+
+  render() {
+    <as-category-widget ref={this.ref} heading={this.props.heading} description={this.props.description}/>
+  }
+}
+```
+
+Then we will list all availiable properties using [React proptypes](https://reactjs.org/docs/typechecking-with-proptypes.html), this tells react
```
ivanmalagon: *available

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,111 @@
...
+    widget.showHeader = this.props.showHeader;
+    widget.showClearButton = this.props.showClearButton;
+    widget.useTotalPercentage = this.props.useTotalPercentage;
+    widget.visibleCategories = this.props.visibleCategories;
+    widget.categories = this.props.categories;
+    widget.addEventListener('categoriesSelected', this._onSelectedChanged.bind(this));
+  }
+
+  _onSelectedChanged(event) {
+    const { onSelectedChanged } = this.props;
```
ivanmalagon: `onSelectedChanged` is not explained.

I get that's a callback prop but it's not listed in propTypes. It's not explained either in the text
`...native listener that connects to '_onSelectedChanged' function. This function will call back the right prop wadus wadus...` You get the point.

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,111 @@
...
+With this steps we created a React wrapper over the native element that allows to pass everything as an attribute
+
+```js
+ <CategoryWidget
+  heading="Business Volume"
+  description="Description"
+  categories={categories}
+  onSelectedChanged={this.onSelectedChanged}
+  showClearButton={this.showClearButton}
+```
```
ivanmalagon: The category widget has also methods that can be hooked up with the wrapping React component. Talk about them also since we have props, events and methods.

IagoLast: eing? I dont follow you

IagoLast:
@ivanmalagon what do you think about the previous comment?

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,115 @@
+This guide contains the basic fundamentals of how to integrate Airhship components with [React.js](https://reactjs.org/).
```
jesusbotella: `Airhship`

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,183 @@
...
+```js
+import { defineCustomElements } from '@carto/airship-components';
+defineCustomElements(window);
+```
+
+In order to manage the state, either listening to events or synchronizing the attributes we got two options.
+
+## Option 1: Manage component logic in the parent layer.
+
+The simplest dirty-hack is to update all attributes, and re-render everything. Even so,  [content attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Content_versus_IDL_attributes) only accept strings as parameters forcing to set the element properties from javascript.
```
ivanmalagon: Remove `dirty`. It sounds like a superñapa.

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,183 @@
...
+
+
+Use the **Render** function you can bind `strings`, `numbers` and `boolean` values as parameters.
```
ivanmalagon: Some suggestions:

In the **Render** function you can bind `string`, `number`and `boolean` values as simple HTML attributes. Just as you do with a normal HTML tag.

In the **componentDidMount** you can setup event listeners and bind attributes that are `arrays` or `objects`. Those type can't be passed as simple HTML attributes so you need to hook them up here.

Use **componentDidUpdate** to update the bind properties values.

docs/guides/10-integrating-react.md
ivanmalagon: There is a `getSelectedCategories ` and a `clearSelection ` method in the category widget.

I don't know how we can expose those methods in the React component but we should.

ivanmalagon: https://shripadk.github.io/react/tips/expose-component-functions.html

It seems that it can be done just using a regular method. The parent should get a ref to the component but it's not our job telling how to use it from the parent, we just have to expose the method in the guide.

docs/guides/10-integrating-react.md
IagoLast: becase it is ;)

jesusbotella: That's not a hack. That's the way to do it :)

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,196 @@
...
+
+> See the full example [here](https://github.com/CartoDB/airship-demos/tree/master/react/widget)
+
+
+First we create a new [React Component]() to represent our widget, in this case a `category-widget`. In the `render` function we will return
+the Airship category widget referenced using a [ref](https://reactjs.org/docs/refs-and-the-dom.html).
+
+
+```js
+export default class App extends Component {
```
jesusbotella: This component must be named `CategoryWidget`

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,196 @@
...
+Then we set up all the bindings in the `componentDidMount` function binding every React property to the corresponding in the web component. We also associate the events, adding a native listener that connects to `_onSelectedChanged` function.
+
+onSelectedChanged is a funcion pased to the component as an attribute and will be called everytime the web components fires an `categoriesSelected` event.
+
+```js
+export default class App extends Component {
+  componentDidMount() {
+    const widget = ref.current;
+
+    widget.showHeader = this.props.showHeader;
```
jesusbotella: Most of these properties can be passed as HTML attributes

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,210 @@
...
+    super(props);
+    this.ref = React.createRef();
+  }
+
+  render() {
+    <as-category-widget 
+      ref={this.ref}
+      heading={this.props.heading}
+      description={this.props.description}
+      show-header="false"
```
jesusbotella: Don't we have to pass these properties based on default props?

docs/guides/10-integrating-react.md
```diff
@@ -50,15 +50,19 @@ In order to manage the state, either listening to events or synchronizing the at
 
 ## Option 1: Manage component logic in the parent layer.
 
-The simplest dirty-hack is to update all attributes, and re-render everything. Even so,  [content attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Content_versus_IDL_attributes) only accept strings as parameters forcing to set the element properties from javascript.
+The simplest way to link Ariship and React is to manually bind attributes and events. [Content attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Content_versus_IDL_attributes) only accept strings as parameters so we need to
```
ivanmalagon: *Airship

docs/guides/10-integrating-react.md
```diff
@@ -50,15 +50,19 @@ In order to manage the state, either listening to events or synchronizing the at
...
+use javascript to add event listeners and pass complex objects as parameter.
 
 
-Use the **Render** function you can bind `strings`, `numbers` and `boolean` values as parameters.
-Use the **componentDidMount** to setup event listeners and bind attributes that you cant bind in the `render` method.
-Use the **componentDidUpdate** to bind properties. 
+In the **render** function you can bind `string`, `number`and `boolean` values as simple HTML attributes. Just as you do with a normal HTML tag.
+
+In the **componentDidMount** you can setup event listeners and bind attributes that are `arrays` or `objects`. Those type can't be passed as simple HTML attributes so you need to hook them up here.
```
ivanmalagon: *those types

docs/guides/10-integrating-react.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Loading web components
+
+Install `airship-components` package 
+
+```
+npm i @carto/airship-components
+```
+
+In the entry point of your application call `defineCustomElements` pasing `window` as a parameter. Once you do this this, all airship components
```
ivanmalagon: passing*

ivanmalagon: *this this <- extra this
Airship must be capitalized.

PR #355, opened 2018-10-01
Remove listeners for Touch events in Range Slider

### LINKED ISSUE #346: Range Slider not working on Mobile
For some reason, once you move a range slider, the touch events won't work anymore.

jesusbotella:
Fix #346.

The `touchstart` and `touchend` event listeners were not being properly removed, hence we were moving all thumbs 🤦‍♂️ .

I updated slider event to make it work in mobile. We were not getting touch event position properly.

### DIFF
--- a/packages/components/src/components/as-range-slider/MouseTrack.ts
+++ b/packages/components/src/components/as-range-slider/MouseTrack.ts
@@ -22,7 +22,11 @@ export class MouseTrack {
 
   private _handleRelease(eventProperties: MouseEvent, listeners: MouseListeners, customListeners: MouseListeners) {
     document.removeEventListener('mousemove', listeners.move);
+    document.removeEventListener('touchmove', listeners.move);
     document.removeEventListener('mouseup', listeners.release);
+    document.removeEventListener('touchend', listeners.release);
+
+    document.removeEventListener('dragstart', this.preventAndStop);
 
     if (customListeners.move) {
       customListeners.move(eventProperties);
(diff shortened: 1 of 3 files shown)


packages/components/src/components/as-range-slider/MouseTrack.ts
```diff
@@ -22,7 +22,11 @@ export class MouseTrack {
 
   private _handleRelease(eventProperties: MouseEvent, listeners: MouseListeners, customListeners: MouseListeners) {
     document.removeEventListener('mousemove', listeners.move);
+    document.removeEventListener('touchmove', listeners.move);
     document.removeEventListener('mouseup', listeners.release);
+    document.removeEventListener('touchend', listeners.release);
+
+    document.removeEventListener('dragstart', this.preventAndStop);
```
IagoLast: why this listener is not in the listeners object?

jesusbotella: Because we don't have any custom references coming from components. It is always added with the same function reference.

packages/components/src/components/as-range-slider/track/as-range-slider-bar.tsx
```diff
@@ -135,7 +135,15 @@ export class RangeSliderBar extends MouseTrack {
   }
 
   private _getMovementDelta(currentEvent: MouseEvent, previousEvent: MouseEvent) {
```
IagoLast:
```ts
private _getMovementDelta(currentEvent: MouseEvent|TouchEvent, previousEvent: MouseEvent|TouchEvent) {
```

PR #365, opened 2018-10-02
Fix img styling in toolbar

### LINKED ISSUE #319: Images / logos in Toolbar size
### Context
Everytime we put a logo or image in the toolbar, we need to add a custom class to avoid the image to show in its original size. Usually, they're huge.

Modify the item element (or add another one for images) to avoid this to happen.

IagoLast:
Fix #319 
---

### DIFF
--- a/packages/styles/src/core/layout/_layout.scss
+++ b/packages/styles/src/core/layout/_layout.scss
@@ -42,6 +42,10 @@
         overflow-x: hidden;
       }
 
+      img {
+        max-width: 100%;
+      }
+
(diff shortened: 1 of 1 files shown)


packages/styles/src/core/layout/_layout.scss
```diff
@@ -42,6 +42,10 @@
         overflow-x: hidden;
       }
 
+      img {
+        max-width: 100%;
```
jesusbotella: Wouldn't it be max-height as well?

PR #366, opened 2018-10-02
Add built-in label to switch component

### LINKED ISSUE #340: Add label to switch component
A common pattern when using the switch component is to add a label next to it.

Add it as a property of the switch so no extra text node nor wrapper is needed.

jesusbotella:
Fix #340

### DIFF
--- a/packages/components/src/components/as-switch/as-switch.html
+++ b/packages/components/src/components/as-switch/as-switch.html
@@ -7,12 +7,33 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0">
   <title>Airship</title>
   <script src="/build/airship.js"></script>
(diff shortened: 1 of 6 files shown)


packages/components/src/components/as-switch/as-switch.html
```diff
@@ -7,13 +7,34 @@
...
+  </style>
 </head>
 
 <body>
   <as-switch id="switch-0" name="switch" title="switch"></as-switch>
   <as-switch checked id="switch-1" name="switch-1" title="switch-1"></as-switch>
   <as-switch disabled id="switch-2" name="switch-2" title="switch-2"></as-switch>
 
+  <div class="with-label">
+    <as-switch checked id="switch-1" name="switch-1" title="switch-1" label="Private Room"></as-switch>
```
IagoLast: duplicated id, name and title

PR #369, opened 2018-10-02
Vue integration guide

### LINKED ISSUE #325: Guides: Integration with Vue
Create a small guide showing how to integrate Airship with Vue

jesusbotella:
Close #325.

### DIFF
--- a/docs/guides/12-Vue-integration.md
+++ b/docs/guides/12-Vue-integration.md
@@ -40,12 +40,12 @@ Although Web Components are supported in Vue, there's one caveat when it comes t
 </as-category-widget>
 ```
 
-To pass complex properties such as Objects and Arrays or listen to component events, you need to do it via Vue element reference or create a mirror component.
+To pass complex properties such as Objects and Arrays or listen to component events, you need to do it via Vue element reference or creating a mirror component.
 
 #### Managing component with element reference
 There are some properties that we cannot pass or events that we cannot listen to in the example above, such as category property or categoriesSelected event.
 
-To get element reference, we need to set `ref` property in the component template node.
+To get the element reference, we need to set `ref` property in the component template node.
 
 ```html
 <template>
@@ -57,7 +57,7 @@ To get element reference, we need to set `ref` property in the component templat
 </template>
 ```
 
-We will find the element's reference in `refs` property in our component scope by doing that, hence will have full access to our component.
+Element's reference will be in our Vue component's `refs` property, hence we will have full access to our Airship component.
 
 ##### Listening to component events
 The best way to listen to component events is in `mounted` function in your desired component.
@@ -75,7 +75,7 @@ export default {
 }
 ```
 
-We create an easy-to-use reference to the widget so that we can reutilize it in other parts of our component. Then, we listen to `categoriesSelected` event where you can put your own custom code to perform the action that you want.
+We create an easy-to-use reference to the widget so that we can reuse it in other parts of our component. Then, we listen to `categoriesSelected` event where you can put your own custom code to perform the action that you want.
 
 It will only listen once to component event and it won't create multiple unneeded event listeners by doing it in `mounted`callback`.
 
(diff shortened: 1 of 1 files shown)


docs/guides/12-Vue-integration.md
```diff
@@ -0,0 +1,197 @@
+This guide will lead you through the process of integrating Airship in [Vue](https://vuejs.org/).
+
+## Including Styles
+Airship styles need to be globally included within our Vue application, so that we can use styles everywhere instead of including them inside any component scope. We are going to import them in `main.js` via npm.
```
IagoLast: to import them in  ~~main.js~~ `entry point` via npm.

docs/guides/12-Vue-integration.md
```diff
@@ -0,0 +1,197 @@
+This guide will lead you through the process of integrating Airship in [Vue](https://vuejs.org/).
+
+## Including Styles
```
ivanmalagon: This works if we have `styles` and `css` loaders, right? Please a note that vue-cli includes them by default but, if not, the user must add them.

docs/guides/12-Vue-integration.md
```diff
@@ -0,0 +1,197 @@
...
+Vue.config.ignoredElements = [/as-\w+/];
+```
+
+Second, we need to import and define our Airship components. As we did before with Airship styles, we import them from npm, and then we need to execute `defineCustomElements` function.
+
+```js
+import { defineCustomElements } from '@carto/airship-components';
+defineCustomElements(window);
+```
+
```
ivanmalagon: Add a note telling that these two steps are better done in the main entry point file.

docs/guides/12-Vue-integration.md
```diff
@@ -0,0 +1,197 @@
...
+    });
+  }
+}
+```
+
+We create an easy-to-use reference to the widget so that we can reuse it in other parts of our component. Then, we listen to `categoriesSelected` event where you can put your own custom code to perform the action that you want.
+
+It will only listen once to component event and it won't create multiple unneeded event listeners by doing it in `mounted`callback`.
+
+##### Passing complex properties to component
```
ivanmalagon: This section appears completely in blue. I don't what's happening. A syntax error?

docs/guides/12-Vue-integration.md
```diff
@@ -0,0 +1,197 @@
...
+    this.categoryWidget.addEventListener('categoriesSelected', (event) => {
+      console.log('Selected Categories', event.detail);
+    });
+  }
+}
+```
+
+We create an easy-to-use reference to the widget so that we can reuse it in other parts of our component. Then, we listen to `categoriesSelected` event where you can put your own custom code to perform the action that you want.
+
+It will only listen once to component event and it won't create multiple unneeded event listeners by doing it in `mounted`callback`.
```
ivanmalagon: Any mention on where to remove the listener? Or if Vue takes care of that automagically, mention that the developer should not worry about it.

docs/guides/12-Vue-integration.md
```diff
@@ -0,0 +1,197 @@
...
+      this.categoryWidget.getSelectedCategories();
+    },
+    clearSelection: function () {
+      this.categoryWidget.clearSelection();
+    }
+  }
+}
+```
+
+### Mirroring Web Components with Vue component
```
ivanmalagon: :+1:

docs/guides/12-Vue-integration.md
jesusbotella: Yep, there is a misplaced ` at the end of the last sentence.

docs/guides/12-Vue-integration.md
jesusbotella: Done!

docs/guides/12-Vue-integration.md
jesusbotella: Any framework should take care of it when destroying the component where the Web Component was placed

PR #370, opened 2018-10-03
Make as-responsive-content docs great again

### LINKED ISSUE #342: as-responsive-content reference fixes
`as-responsive-content` section has several issues:
- it's not clear where it should be placed. It has to replace the `as-content` node but it's not stated anywhere and it's super important.
- `ready` event is also key to using this component. It should be above its current place, in the first places of the section.

jesusbotella:
Fix #342

### DIFF
--- a/packages/components/src/components/as-responsive-content/README.md
+++ b/packages/components/src/components/as-responsive-content/README.md
@@ -14,6 +14,38 @@ The component searches for these classes in order to know how many tabs it must 
 
 As long as you follow the proper markup, as the `Layout` section shows, you won't have to worry about anything else.
 
+This component replaces `.as-content` HTML element, so it needs to wrap all your children elements to work properly. Here is an example:
+
+```code
+lang: html
+---
+<body class="as-app">
+  <!-- Here goes your toolbar -->
+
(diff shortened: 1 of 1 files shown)


packages/components/src/components/as-responsive-content/README.md
```diff
@@ -14,6 +14,38 @@ The component searches for these classes in order to know how many tabs it must
 
 As long as you follow the proper markup, as the `Layout` section shows, you won't have to worry about anything else.
 
+This component replaces `.as-content` HTML element, so it needs to wrap all your children elements to work properly. Here is an example:
```
IagoLast: I think this should be with the `>` markdown tag, this way its shown as a ⚠️  in our docs.

> This component replaces `.as-content` HTML element, so it needs to wrap all your children elements to work properly.

jesusbotella: I don't know if we should put it like a warning 🤔 
It should be clear as long as you read the whole text. I put the ready event text as a warning, because I felt like it was more important.

ivanmalagon:
LGTM also.

PR #371, opened 2018-10-03
Point the layout placement

### LINKED ISSUE #339: Add in layout reference a mention to the nodes
We need to add a note about how the different nodes have to be in layout. We have a visual guide but in the user testing we found him inserting elements where they shouldn't be (the toolbar inside content). Perhaps a mention of the node hierarchy in the document can help in this sense.

IagoLast:
Fix #339 
---


ivanmalagon:
LGTM

PR #381, opened 2018-10-03
Add Switch label documentation

### LINKED ISSUE #376: Add switch label to documentation
Switch component has now a label property but there's no mention of it in the reference.

jesusbotella:
Fix #376

### DIFF
--- a/packages/components/src/components/as-switch/README.md
+++ b/packages/components/src/components/as-switch/README.md
@@ -4,7 +4,7 @@ It is a two state button that can be on or off. This component issues a "changed
 
 ## Examples
 
-### Simple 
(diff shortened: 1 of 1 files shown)


packages/components/src/components/as-switch/README.md
```diff
@@ -67,9 +72,23 @@ lang: javascript
 switch.name = 'toggle-layer';
 ```
 
+### **label**: boolean
+Text for label. Positioned on the right of the switch.
```
IagoLast: Boolean? 🤔 

And about the description:
> When this attribute is set a label will appear on the right of the switch with the given value as text.

PR #395, opened 2018-10-05
Set proper encoding when reading files to upload

jesusbotella:
I noticed the other day that a character we introduced in the BANNER file was not being properly shown in S3 file.

I think that was because the uploaded files were not being read with the proper encoding, so this PR fixes that, hopefully.

screen shot 2018-10-05 at 12 36 42


IagoLast:
cc @Jesus89, this might interest you :)

Issue #398, opened 2018-10-05
Release Beta version of Airship


ivanmalagon:
Upgrade Airship to beta status and ship it.

- [x] Update CHANGELOG
- [x] Adapt README
- [x] Update version
- [x] Publish to NPM
- [x] Publish to CDN
- [x] Change Airship minisite link to documentation
- [x] Profit

ivanmalagon:
🚀

PR #403, opened 2018-10-09
Adapt README to public release

ivanmalagon:
Related to #398
---


README.md
```diff
@@ -14,15 +15,23 @@ There are three npm packages for easier use:
 
 ## Documentation
```
jesusbotella: I'd change the order to put `What is Airship?` first, and then the documentation link.

README.md
```diff
@@ -1,4 +1,5 @@
-# Airship
+# ![](airship.svg) Airship
```
jesusbotella: What about removing `There are three npm packages for easier use:` or changing it to make it clearer?

The logo is too big compared to the text. What about using a `` tag and set the height to the font size?

PR #406, opened 2018-10-09
Update Stencil to v0.13

jesusbotella:
This PR updates StencilJS to v0.13.2. Implementing some breaking changes:

- **Component Method invocation is now handled by Promises**
  Stencil changed Method specification to return a Promise which holds the value of the method invokation. Given that we told developers to use component methods in our documentation, we should change that to let them know the change.

- **DOM rendering in testing is now performed by Puppeteer instead of JSDom**
   Testing has completely changed from previous version to this one. It's more powerful and easier to use because of Puppeteer.

Please read more about them [here](https://github.com/ionic-team/stencil/blob/master/CHANGELOG.md#-0130-2018-09-26).


jesusbotella:
Please wait until documentation is changed to reflect Stencil Breaking changes.

PR #410, opened 2018-10-10
Update documentation to reflect @Method changes in Stencil 0.13

jesusbotella:
Update documentation accordingly to #406 and StencilJS breaking changes.

### DIFF
--- a/packages/components/src/components/as-category-widget/README.md
+++ b/packages/components/src/components/as-category-widget/README.md
@@ -240,13 +240,23 @@ categoryWidget.addEventListener('categoriesSelected', event => {
 
 #### **getSelectedCategories**
 Get current selected categories
-`Returns: Category[]`
+`Returns: Promise<Category[]>`
 
 ```code
 lang: javascript
 ---
+// Async/Await approach
+// You need to wrap your code in an async function in top level scripts
+(async function () {
+    const categoryWidget = document.querySelector('as-category-widget');
+    await categoryWidget.getSelectedCategories();
+})();
+
+
+// Promises approach
 const categoryWidget = document.querySelector('as-category-widget');
-categoryWidget.getSelectedCategories();
+categoryWidget.getSelectedCategories()
+              .then(categories => console.log(categories));
 ```
 
 #### **clearSelection**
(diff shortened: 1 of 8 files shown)


packages/components/src/components/as-category-widget/README.md
```diff
@@ -240,13 +240,23 @@ categoryWidget.addEventListener('categoriesSelected', event => {
...
 ```code
 lang: javascript
 ---
+// Async/Await approach
+// You need to wrap your code in an async function in top level scripts
+(async function () {
+    const categoryWidget = document.querySelector('as-category-widget');
+    await categoryWidget.getSelectedCategories();
+})();
+
```
ivanmalagon: I'd put the simple code (no IIFE) and then a note telling that if you need to use it in a top level script you need to wrap, with a link to a deeper explanation.

packages/components/src/components/as-category-widget/README.md
```diff
@@ -255,8 +265,20 @@ Clear current selected categories
 ```code
 lang: javascript
 ---
+// Async/Await approach
+// You need to wrap your code in an async function in top level scripts
+(async function () {
+    const categoryWidget = document.querySelector('as-category-widget');
+    await categoryWidget.clearSelection();
+})();
```
ivanmalagon: Same as above

packages/components/src/components/as-category-widget/README.md
```diff
@@ -240,13 +240,24 @@ categoryWidget.addEventListener('categoriesSelected', event => {
...
+
+// Promises approach
 const categoryWidget = document.querySelector('as-category-widget');
-categoryWidget.getSelectedCategories();
+categoryWidget.getSelectedCategories()
+              .then(categories => console.log(categories));
+```
+
+```hint|directive
+Please note that you always need to wrap your `await` code in an `async` function. If you use it outisde of an async function, it will raise a `SyntaxError`. Learn more about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await).
```
ivanmalagon: `outisde` instead of `outside` all over the place.

PR #414, opened 2018-10-17
Add body class

### LINKED ISSUE #409: Nodes of components in Vue
This is a placeholder to enter information about how the Airship styles behave when inserted in a Vue application. As Vue doesn't have fragments as of now, a component needs to be wrapped in only one node.

For the experience of @Jesus89 this makes that some styles don't work properly and he needs to set the style for those nodes.

Check what happens for real and provide a solution that can be useful for other frameworks as well.

ivanmalagon:
Fixes #409
---

After doing a POC using Vue to see where it's mandatory to break our cascade when using components, I've come to the conclusion that our main problem was the `as-app` selector.

In all of our examples `as-app` goes in the `body` tag and up to three nodes hang from it: `as-toolbar`, `as-tabs` and `as-content`. That causes that the main app component needs an enclosing `div` that breaks the flex cascade from body.

Splitting the old `as-app` into `as-app-body` and `as-app` allows to use `as-app` in the enclosing tag, not breaking anything. And it allows that the Airship app can be along with other content in the page.

Note: I chose `as-app-body` because `as-body` is already taken by the typography classes.

Note': changes in documentation still pending.

### DIFF
--- a/docs/guides/01-introduction.md
+++ b/docs/guides/01-introduction.md
@@ -31,29 +31,31 @@ The easiest way to include Airship is through our CDN, adding the tags in the he
   <!-- Include Mapbox GL CSS -->
   <link href="https://libs.cartocdn.com/mapbox-gl/v0.48.0-carto1/mapbox-gl.css" rel="stylesheet" />
 </head>
-<body class="as-app">
+<body class="as-app-body">
 
-  <header class="as-toolbar"></header>
-  <nav class="as-tabs"></nav>
+  <div class="as-app">
+    <header class="as-toolbar"></header>
+    <nav class="as-tabs"></nav>
 
-  <div class="as-content">
-    <aside class="as-sidebar as-sidebar--left"></aside>
+    <div class="as-content">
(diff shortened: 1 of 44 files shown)


IagoLast:
I dont understand this, now we have to add 2 classes into the `body` tag? can you put an example where is used differently?

ivanmalagon:
It's not needed necessarily in the body tag.

I'm rewriting the docs to turn these snippets
```html
  <body class="as-app as-app--nav-left">
    <header class="as-toolbar"></header>
    <div class="as-content"></div>
  </body>
```

into

```html
  <body class="as-app-body">
    <div class="as-app as-app--nav-left">
      <header class="as-toolbar"></header>
      <div class="as-content"></div>
   </div>
  </body>
```

This way:
- an `App` component can export a single node containing all the content and we don't break the flexbox chain.
- the `nav-left` stuff stays at a container level, just above the actual toolbar, instead of setting it at `body` level, that doesn't feel right.

Apart from that, do you have any other suggestion to `as-app-body`?

Issue #415, opened 2018-10-17
Active state should have its own css variable


rjimenezda:
# The :bug: part
While customizing my Airship app with my own colors I noticed that you're changing the background color of buttons on hover and active with the same variable but different fallbacks (darken 20%, darken 40%). Like this:

buttons.scss
```scss
a.as-btn.as-btn--primary {
  &:hover {
    background-color: var(--as-button-primary-color-focus, darken($color-primary, 20%));
  }

  &:active {
    background-color: var(--as-button-primary-color-focus, darken($color-primary, 40%));
  }
}
```

I think this makes it impossible to customize both states using variables without overriding CSS. 

I would propose changing `--as-button-primary-color-focus` to `--as-button-primary-color-hover` and adding `--as-button-primary-color-active`. Since that's a breaking change, maybe just adding the new one is fine, and keeping the old one for `:hover`

This is not the only example, though. There's a few more cases on the same file.

# The :question: part
I also noticed that the secondary button does not have an active state. Or rather, it's exactly the same as the hover. Is this intended?

rjimenezda:
Oh, that's interesting. The secondary button does have it's own variable (`--as-button-secondary-color-active`), but the default fallback is the same. The question still stands, though, is that intended?

rjimenezda:
Closed via #416

PR #416, opened 2018-10-18
Fix button active state variables

rjimenezda:
#415 
---

I've added new variables for active states where they were missing.

Also, i've added a darken 12% to secondary buttons


rjimenezda:
Also fixed the hrefs on the HTML elements docs, they had '"#' and broke when clicking on them

PR #426, opened 2018-10-23
Change z-index to 0

elenatorro:
While using the `<as-range-slider>` component, I ran into the issue of not having the back rail bar visible. This is because the back rail bar has a z-index level of -1. I **think** it is not necessary. When using `z-index: -1`, it is set backwards by default, and therefore, it is not visible.

Example:

screen shot 2018-10-23 at 11 23 03
screen shot 2018-10-23 at 11 25 23

### DIFF
--- a/packages/components/src/components/as-range-slider/slider/as-range-slider.scss
+++ b/packages/components/src/components/as-range-slider/slider/as-range-slider.scss
@@ -21,7 +21,7 @@ as-range-slider {
     &::before {
       content: ' ';
       position: absolute;
-      z-index: -1;
+      z-index: 0;
       top: 50%;
       left: 0;
       width: 100%;


IagoLast:
LGTM but I think this was there for a reason...

@jesusbotella @VictorVelarde any idea?

elenatorro:
It is not critical, the user can change the `z-index` of the wrapper, **but** I think it may confuse non experience users because it is kind of tricky to figure out you need to modify the `z-index` since slider is actually visible.

jesusbotella:
I think that @albertoac told me the same issue. I don't know why it was there but I think it needs to be changed.

IagoLast:
so lets merge it :)

ivanmalagon:
This PR should have a commit with the conventional commit format to reflect this change in the changelog.

Please add an empty commit in `master` with the right format.

`fix(components): change as-range-slider rail z-index to 0 #426`

VictorVelarde:
If conventional commit is being followed, then IMO it should be required. https://github.com/Kikobeats/automate-release first part on commitlint can be used as a reference

ivanmalagon:
First of all, we don't use stuff from @Kikobeats here.

Second, I'm just kidding. Didn't know about commitlint.

Kikobeats:
I can give a talk for food

IagoLast:
👍 for commitlint

PR #431, opened 2018-10-23
Change publish script

ivanmalagon:
This PR fixes the latest broken builds.

I haven't moved the uploaded folder from `www/build` to `dist` until we have a response from Ionic about the best way to do it.

### DIFF
--- a/packages/components/stencil.config.js
+++ b/packages/components/stencil.config.js
@@ -10,7 +10,6 @@ exports.config = {
     {
       type: 'www',
       serviceWorker: false,
-      empty: true,
       dir: path.join(__dirname, '../../www')
     }
   ],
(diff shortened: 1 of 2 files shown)


packages/components/stencil.config.js
```diff
@@ -10,7 +10,7 @@ exports.config = {
     {
       type: 'www',
       serviceWorker: false,
-      empty: false,
+      empty: true,
```
IagoLast: true is the default value so we can remove this line :)

ivanmalagon:
For the record:

we had `test` after `build` in the publish script. Prior to Stencil 0.13, it didn't matter because `test` was not creating any assets. After 0.13, it creates the dev assets so we were publishing the development ones.

PR #437, opened 2018-10-25
Add serve to npm run dev

ivanmalagon:
I missed the local server when running `npm run dev`. It was a change in Stencil and now you have to pass another option to start the server.

For me, it's very convenient to have the server running. I know we use from time to time `npm run showcase` that serves the files as well but I think it should be available in `dev`.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -11,6 +11,12 @@
 
 This is a [Lerna](https://lernajs.io/) [monorepo](https://en.wikipedia.org/wiki/Monorepo) composed of 3 node packages in the `packages`.
 
+## Developing
+
+Run `npm run dev` to start developing. This commands starts the Stencil compiler for the components and the Sass compiler for the styles. Both in watch mode.
+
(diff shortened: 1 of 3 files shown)


IagoLast:
try running `npm run serve` instead.

ivanmalagon:
Is there any good reason to have to run a server explicitly when developing?

If there's one, then I'll add it to `DEVELOPERS.md`.

If not, then we can merge this PR that adds `& npm run serve` to the general `dev` script.

ivanmalagon:
Forget about it.

`&` is not a good option since the Stencil compiling gets running when pressing `Ctrl + C`.

Adding that clue to DEVELOPERS.md

DEVELOPERS.md
```diff
@@ -11,6 +11,12 @@
 
 This is a [Lerna](https://lernajs.io/) [monorepo](https://en.wikipedia.org/wiki/Monorepo) composed of 3 node packages in the `packages`.
 
+## Developing
```
IagoLast: Add this to the index and 🚀

PR #442, opened 2018-10-26
217 histogram redesign

### LINKED ISSUE #217: Histogram redesign
There are several issues to improve in the current histogram implementation, mostly related to range selection.

**1. Redesign selection**
image

The two balls seemed a good solution but it's not visually appealing depending on the number of buckets. In the picture above we can see that the balls don't get along with the buckets. If you choose less bucket it's even worse.

**2. Range values**

We don't know the values of the selected range. We miss the visual clue of the extremes that we have in Builder. We can add a text telling what they are as we do in category widgets or do something more sophisticated like in Builder.

**3. Click on a bucket**

Right now, you have to drag the mouse in order to select only one bucket. What the user expects is to click on a single bucket and get it filtered, like it happens in the category widget.

IagoLast:
Fix #217
---


packages/components/src/components/as-histogram-widget/as-histogram-widget.tsx
```diff
@@ -27,7 +28,7 @@ const MARGIN = {
   TOP: 5,
   YAxis: 20
 };
-const CUSTOM_HANDLE_Y_COORD = HEIGHT + MARGIN.TOP - (CUSTOM_HANDLE_SIZE / 2);
+const CUSTOM_HANDLE_Y_COORD = HEIGHT + MARGIN.TOP - (28 / 2);
```
rjimenezda: 28 => Magic number?

packages/components/src/components/as-histogram-widget/utils/data.service.ts
```diff
@@ -0,0 +1,26 @@
+import { max } from 'd3-array';
```
rjimenezda: why data.service.ts and not data-service.ts?

packages/components/src/components/common/as-widget-header/as-widget-header.tsx
```diff
@@ -30,10 +30,38 @@ export class WidgetHeader {
...
+
+  private _getSubHeader() {
+    if (this.isLoading) {
+      return <p class='as-widget-header__subheader as-body '>{this.subheader}</p>;
+    }
+    if (this.error) {
+      return <p class='as-widget-header__subheader as-widget-header__subheader--error as-body '>{this.error}</p>;
+    }
+    if (this.isEmpty) {
+      return <p class='as-widget-header__subheader as-widget-header__subheader--empty as-body '>NO DATA AVAILABLE</p>;
```
rjimenezda: Did we ever settle on I18N issues? This is one of them.

I don't think a component library should have any string literals like this one.

packages/components/src/components/common/content.fragment.tsx
```diff
@@ -0,0 +1,22 @@
...
+ */
+export default function(isLoading, error, isEmpty, heading, errorDescription, fragment) {
+  if (isLoading) {
+    return <as-loader class={heading ? 'content as-pb--36' : 'content as-pb--20'}></as-loader>;
+  }
+  if (error) {
+    return <p class='content as-body'>{errorDescription || 'Unexpected error'}</p>;
+  }
+  if (isEmpty) {
+    return <p class='content as-body'>There is no data to display.</p>;
```
rjimenezda: Same as above, we should provide at least an argument to customize this.

packages/components/src/components/as-histogram-widget/utils/data.service.ts
rjimenezda: Same with the fragment thingie

IagoLast: - Names are separated with hyphens -> semantic meaning
- Separate names with periods -> indicate the type of file (test, service, fragment ...)

rjimenezda: :+1: 

Now that I think about it, I did the same with fixtures on builder :eyes:

PR #446, opened 2018-10-30
Use steps when moving the range slider thumb

### LINKED ISSUE #429: Range slider steps are not working
### Context

When creating a `<as-range-slider>` component and setting the `step` value, it is not applied to the component

### Steps to Reproduce

1. Add `<as-range-slider id="simple" step="5" min-value="5" max-value="30"></as-range-slider>`
2. Drag the slider

### Current Result

It does not move the slider 5 by 5, but 1 by 1 instead

### Expected result

It should apply the `step` value properly

### Browser and version

Chrome

rjimenezda:
Fix #429
---

It was working properly for keyboard events and moving the bar, but not for moving the thumb.

### DIFF
--- a/packages/components/src/components/as-range-slider/slider/as-range-slider.tsx
+++ b/packages/components/src/components/as-range-slider/slider/as-range-slider.tsx
@@ -228,7 +228,7 @@ export class RangeSlider {
     const isRightThumb = rightThumb === thumb;
 
     const value = this._getValueFromPercentage(percentage);
-    const stepValue = this._getStepValue(value);
+    const stepValue = this._getStepValueWithThreshold(value, thumb.value);
     const stepPercentage = this._getPercentage(stepValue);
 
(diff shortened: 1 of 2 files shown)


IagoLast:
(update tests)

rjimenezda:
They actually are, but there were no specs for steps (I'm on it though)

elenatorro:
💖

packages/components/src/components/as-range-slider/slider/as-range-slider.tsx
```diff
@@ -304,23 +304,10 @@ export class RangeSlider {
   }
 
   private _getStepValue(value) {
-    const stepValue = (value / this.step) * this.step;
+    const stepValue = Math.round(value / this.step) * this.step;
```
VictorVelarde: 👍 great catch! @rjimenezda

PR #448, opened 2018-10-30
Dropdown issues

### LINKED ISSUE #433: Set min-height for dropdown control
### Context

https://carto.com/developers/airship/examples/#example-as-dropdown

### Steps to Reproduce

1. Create an empty dropdown with default-text=""

### Current Result

selection_357

### Expected result

selection_358

```css
as-dropdown .as-dropdown__control {
  min-height: 36px;
}
```

### LINKED ISSUE #447: Dropdown remains open after clicking outside
**What is happening?**
Using the dropdown component, if you open it and click outside of it, it remains open.

**What should happen?**
The dropdown should close if you click anywhere else

**Steps to Reproduce**
https://carto.com/developers/airship/examples/#example-as-dropdown

Open it, click outside

**Device/Browser/Airship-Version**
N/A

**Extra information**
N/A

rjimenezda:
Fix #433 
Fix #447 
---

Not in love with having to bind, but it doesn't work otherwise (neither the functionality nor clearing the event on unload).


packages/components/src/components/as-dropdown/as-dropdown.e2e.ts
```diff
@@ -98,6 +98,24 @@ describe('as-category-widget', async () => {
...
+
+      await page.waitForChanges();
+
+      await page.click('.as-dropdown__control');
+      await page.waitForChanges();
+
+      const menuDropdown = await page.find('.as-dropdown');
+      expect(menuDropdown).toHaveClass('as-dropdown--open');
+
+      await page.click('body');
```
IagoLast: does this work if we click on some other element? (the footer, another button..)

rjimenezda: It should, unless that particular element stops the propagation (like we do on the component).

IagoLast: 🚀

packages/components/src/components/as-dropdown/as-dropdown.tsx
```diff
@@ -55,6 +55,7 @@ export class Dropdown {
 
   @State() private isOpen: boolean = false;
   @State() private selectedOptionObject: DropdownOption = {};
+  private _docListener: () => void;
```
IagoLast: I think we can use closeList as the default value

rjimenezda: Hmmm I dunno, let me try it. There was a context issue, hence the bind call below.

rjimenezda: It works, but only if I bind (as expected). `private _docListener: () => void = this._closeList.bind(this)`

So would you propose making this a Prop() users can configure?

packages/components/src/components/as-dropdown/as-dropdown.tsx
```diff
@@ -45,6 +45,15 @@ export class Dropdown {
...
   @Prop() public showClearButton: boolean = false;
 
+  /**
+   * Function called when clicking outside of the dropdown.
+   * By default it closes the list.
+   *
+   * @type {function}
+   * @memberof Dropdown
+   */
+  @Prop() public onClickOutside: () => void = this.closeList.bind(this);
```
IagoLast:
```suggestion
  @Prop() 
  public onClickOutside: () => { 
      this._closeList();
  }
```

??

I think this works out of the box, there is no need for the public method `closeList` because if you want to change the default behaviour you just need to override this:

```js
dropdopn.onClickOutside = () => alert('foo');
```

packages/components/src/components/as-dropdown/as-dropdown.tsx
```diff
@@ -56,6 +65,16 @@ export class Dropdown {
   @State() private isOpen: boolean = false;
   @State() private selectedOptionObject: DropdownOption = {};
 
+  /**
+   * Closes the list, useful in case you need to customize {onClickOutside}
+   *
+   * @memberof Dropdown
+   */
+  @Method()
+  public async closeList() {
```
IagoLast: I think this is not needed

rjimenezda: If you override onClickOutside, the dropdown will not close anymore, because we're removing the listener when it changes. 

This method is a way for the user to close the list whenever they think is necessary, for example, after doing something on their custom onClickOutside.

packages/components/src/components/as-dropdown/as-dropdown.tsx
rjimenezda: It's my fault for accepting the suggested change, but it was completely broken syntax-wise

packages/components/src/components/as-dropdown/as-dropdown.tsx
ivanmalagon: This new method is public but it's not documented in the reference.

PR #462, opened 2018-11-05
Fix toolbar actions taking too much space

### LINKED ISSUE #413: Toolbar actions taking a lot of space
### Current result

Having three nodes in a toolbar , one of them being `as-toolbar__actions`. This actions node takes all the available space in the container, pushing itself to the right

```html
  <header class="as-toolbar">
    <div class="as-toolbar__item"></div>
    <nav class="as-toolbar__actions"></nav>
    <div class="as-toolbar__item"></div>
  </header>
```

image

image

### Expected result

`as-toolbar__actions` takes the needed space and all toolbar nodes layout as expected, as `space-between`.

image

ivanmalagon:
Fixes #413 
---


packages/components/src/components/as-toolbar/as-toolbar.e2e.ts
```diff
@@ -39,9 +39,12 @@ describe('as-toolbar', () => {
...
       await page.waitForChanges();
 
-      const actual = await page.find('.as-toolbar__actions');
+      const actual = await page.waitForSelector('.as-toolbar__actions--visible', {
+        timeout: 500,
+        visible: true
+      });
 
-      expect(actual).toHaveClass('as-toolbar__actions--visible');
+      expect(actual).toBeDefined();
```
IagoLast: Does this test fails when not clicking on .as-toolbar__toggle ?

ivanmalagon: Yes. I've checked that it fails when you don't click on the toggle or when `as-toolbar__actions--visible` is not available.

IagoLast: what does the method return? can we use `.toBeThruty` instead?

ivanmalagon:
1. It's useful since it checks that when the toggle is clicked, the right selector is added to the actions.

2. You can check the new screenshot references for this PR. Pull the branch and take a look on them. I don't get the logo CSS comment. The original issue shows that whenever an actions node is present, the expected alignment is not respected.
I've checked that mobile version didn't get affected by these changes.

packages/components/src/components/as-toolbar/as-toolbar.e2e.ts
ivanmalagon: That was my first intuition, using `.toBeTruthy` but I checked that `toBeDefined` has the desired outcome.

According to the Puppeteer API, `waitForSelector` returns an element if it's found. If it's not found, it throws an error. So this `expect` is there only to have a green check.

I'm refactoring it to reflect better the API and expect no errors (that it's the best thing to do).

ivanmalagon: It's not straightforward to check `not.toThrow` and it leads to false positives.

Just changed to `.not.toBeNull` and checked all the combinations.

PR #466, opened 2018-11-06
454 add responsive option to widgets

### LINKED ISSUE #454: Add responsive option to widgets
**Problem**
When the viewport size changes the widget keep their original size leading to bad UX.

**Proposed solution**
Add a new `responsive` boolean parameter to all the widgets. When this flag is set to true, the widget will be re-rendered when the [resize](https://developer.mozilla.org/en-US/docs/Web/Events/resize) event is fired.

```html
<as-widget responsive>
</as-widget>
```

IagoLast:
Fix #454 
---

- [x] stacked bar
- [ ] histogram
- [ ] category

---

### DIFF
--- a/packages/components/src/components.d.ts
+++ b/packages/components/src/components.d.ts
@@ -272,6 +272,10 @@ export namespace Components {
     */
     'noDataHeaderMessage': string;
     /**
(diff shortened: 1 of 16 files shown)


IagoLast:
This will be better splited in 2 different PR

- See this for stacked bar https://github.com/CartoDB/airship/pull/467

Issue #150, opened 2018-08-10
Refactor typography names


ivanmalagon:
Our first approach to create text color utilities used a numeric serie, just like in the Sketch design:

![image](https://user-images.githubusercontent.com/1078228/43950898-f29bd060-9c91-11e8-840d-7d8246e75fe6.png)

**Pros:**
- Independent from theming. Once you know `01` means the main color and the rest is a scale to the complementary one you can implement a dark theme without worrying about semantics.

**Cons:**
- Non semantic. You need to know what `01` to `04` means. Hard to read if you don't.

After talking with @piensaenpixel we thought we should introduce a more semantic naming. We've tried with different combinations and come up with this solution:

![image](https://user-images.githubusercontent.com/1078228/43951076-6d287388-9c92-11e8-962d-9969600d79a8.png)

_Same naming for a light and a dark theme_

The extremes are called `type-primary` and `type-complementary`. In a light scheme, this translates to black and white. In a dark one, to white and black. If you change theme, you won't have to change markup.

We could find any meaningful name for those values in between that works for light and dark themes. For those, we've chosen the first approach, a numeric one.

What do you think @CartoDB/frontend-team ? Any thoughts?

VictorVelarde:
It’s hard to find something semantic, but I think it is worthy. I’ll try to find something related to nearest colors, with an idea of “graduation” eg. something like `type-primary > type-primary-plus > type-secondary-minus > type-secondary`

rubenmoya:
I don't think you can use the graduation approach since the user can set totally different colors, being `primary` red and `primary-plus` dark gray, which it's completely different.

To be honest, I don't see any problem using `$color-[category]-[number]`.

The approach Ivan showed still has the same problem, but now with a mix of semantic naming + numeric naming.

To me the solution would be to keep it as it is, and let the user make it semantic however they want, for example:

```
$color-type-title: $color-type-01;
$color-type-header: $color-type-02;
$color-type-red: $color-type-03;
$color-type-blue: $color-type-04;
```

We should just provide the 4 colors, and let the users decide if they want to use the numeric naming or build something more semantic on top.

VictorVelarde:
IMO that's an important point: is there a graduation idea on those colors or not?

piensaenpixel:
vamos ya

PR #470, opened 2018-11-08
Add css variables rules proposal

### LINKED ISSUE #455: Formal definition about how CSS variables work with Airship.
- [x] Review and list all variables
- [x] Define naming rules (BEM?)
- [x] Update developers.md
- [x] Create a interactive acceptance example
- [ ] Update styling guides

IagoLast:
Fix #455 
---

Tasks:

- [x] Add CSS variables DOC
- [x] Add css variables example/smoke
- [x] Review and update all css variables
- [ ] ~~Update css variables docs~~ (I think we should create docs in a different issue, maybe we want to QA this internally first)

Review https://github.com/CartoDB/airship/issues/150

@ivanmalagon  @rjimenezda you can take a look. Tons of changes here but summarizing:

- All variables have a common naming.
- Variables are linked so they always have a valid fallback.
  - Due to this, some fallbacks were removed to reduce codebase-size.
- Some elements have custom variables and other ones can be extended in the future while keeping backward compatibility.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -18,6 +18,7 @@
     - [Naming](#naming)
       - [Global css-values](#global-css-values)
       - [Global css-variables](#global-css-variables)
+      - [Specific css-values](#specific-css-values)
       - [Specific css-variables](#specific-css-variables)
 
 
@@ -177,6 +178,7 @@ We define two kinds of variables, global and specific:
 
 - **Global Values:** This variables are declared in `styles/core.scss` under the `:host` selector and define specific values that are availiable on every css scope.
 - **Global Variables:** This variables are declared in `styles/core.scss` under the `:host` selector and are availiable on every css scope.
+- **Specific Values:** This variables are declared in the element/component styles and define a value that can be used in different places, usually specific variables. (Same value could be used in the table border and paragraph color)
 - **Specific Variables:** This variables are defined in the stylss of element/component and they only are availiable under the element/component scope.
 
 The variables should have a less specific one as fallback, so for example when developing the `histogram` widget:
@@ -187,7 +189,7 @@ The variables should have a less specific one as fallback, so for example when d
   --as--color--primary: white;
 
   /** <stacked-bar-widget.scss> The values used with chained fallbacks, last fallback is a css-value **/
-  --stacked-bar-widget--background-color: var(--as--widget__header--background-color, --as--color--primary);
+  --as-stacked-bar-widget--background-color: var(--as--widget__header--background-color, --as--color--primary);
 ```
 
 Doing this users can controll specific widgets `--stacked-bar-widget--background-color` share a style for all widgets `--as--widget__header--background-color` or just redefine the default value `--as--color--primary`.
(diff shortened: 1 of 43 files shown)


DEVELOPERS.md
```diff
@@ -162,4 +168,67 @@ Once a new version is released just run `npm run changelog` and push your change
 
 We dont want prereleases to be shown in the changelog so **git tags corresponding to a prerelase should be removed once the QA is done**.
 
-> NOTE: You only need to remove it from github and the `changelog` script will prune all removed tags.
\ No newline at end of file
+> NOTE: You only need to remove it from github and the `changelog` script will prune all removed tags.
+
+## CSS Variables
+
+To achieve a better customization we use [css variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables). Follow this rules in order to write them in a consistent way.
```
rjimenezda:
```suggestion
To achieve a better customization we use [css variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables). Follow these rules in order to write them in a consistent way.
```

DEVELOPERS.md
```diff
@@ -162,4 +168,67 @@ Once a new version is released just run `npm run changelog` and push your change
...
+
+We define two kinds of variables, global and specific:
+
+- **Global Values:** This variables are declared in `styles/core.scss` under the `:host` selector and define specific values that are availiable on every css scope.
```
rjimenezda:
```suggestion
- **Global Values:** This variables are declared in `styles/core.scss` under the `:root` selector and define specific values that are available on every css scope.
```

DEVELOPERS.md
```diff
@@ -162,4 +168,67 @@ Once a new version is released just run `npm run changelog` and push your change
...
+- **Global Variables:** This variables are declared in `styles/core.scss` under the `:host` selector and are availiable on every css scope.
```
rjimenezda:
```suggestion
- **Global Variables:** This variables are declared in `styles/core.scss` under the `:root` selector and are available on every css scope.
```

DEVELOPERS.md
```diff
@@ -162,4 +168,67 @@ Once a new version is released just run `npm run changelog` and push your change
...
+- **Specific Values:** This variables are declared in the element/component styles and define a value that can be used in different places, usually specific variables. (Same value could be used in the table border and paragraph color)
+- **Specific Variables:** This variables are defined in the stylss of element/component and they only are availiable under the element/component scope.
```
rjimenezda:
```suggestion
- **Specific Variables:** This variables are defined in the styles of element/component and they're only  available under the element/component scope.
```

DEVELOPERS.md
```diff
@@ -162,4 +168,67 @@ Once a new version is released just run `npm run changelog` and push your change
...
+
+The variables should have a less specific one as fallback, so for example when developing the `histogram` widget:
```
rjimenezda:
```suggestion
The variables should have a less specific one as fallback, so for example when developing the `stacked-bar` widget:
```

DEVELOPERS.md
```diff
@@ -162,4 +168,67 @@ Once a new version is released just run `npm run changelog` and push your change
...
+```css
+
+  /** <style/core/core.scss> The values are initialized **/
+  --as--color--primary: white;
+
+  /** <stacked-bar-widget.scss> The values used with chained fallbacks, last fallback is a css-value **/
+  --as-stacked-bar-widget--background-color: var(--as--widget__header--background-color, --as--color--primary);
+```
+
+Doing this users can controll specific widgets `--stacked-bar-widget--background-color` share a style for all widgets `--as--widget__header--background-color` or just redefine the default value `--as--color--primary`.
```
rjimenezda: This paragraph is confusing, I rewrote it but I don't know if it's exactly what you meant.

```suggestion
With `--stacked-bar-widget--background-color` users can control specific widgets, with the shared variable `--as--widget__header--background-color` they can control all widgets, or they can just redefine the default value `--as--color--primary`.
```

By the way, the class is `--widget-background-color` right now, not `--as--widget__header--background-color`. I don't know if you plan to change it, just noticed it.

IagoLast: my idea is to change them to follow the rules :)

rjimenezda: great

PR #472, opened 2018-11-08
Responsive histogram

rjimenezda:
This PR rewrites the histogram to be more in line with the stacked bar.

As it is, is 100% functional (as in: everything works), I think we could move a lot of things outside the component class now. Might do that on this PR next.

I've purposefully kept its own y-axis function instead of reusing the one that the stacked bar uses, because I think the goal should be having a generic ChartComponent / AxisComponent / D3Component that shares features with this one and the Stacked bar.

I've only touched the stackedbar to wrap the rendering inside of requestAnimationFrame. The reason for this is that whenever a property (i.e: heading) changes, if the layout changes and the SVG has less space, the rendering has to take place after the current paint cycle is done, so that the dimensions are updated.

Also, since the histogram has animations, I've added a flag to not have them when resizing because they look weird. Plus, I've made it so that the animation delay is based on the bar position, I've wanted to do that for a while but never got around to it. I can try making them similar in the stacked bar, if there's time :grimacing:

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -11,6 +11,8 @@
   - [Releasing a new version](#releasing-a-new-version)
     - [Prerelase](#prerelase)
     - [Release](#release)
+    - [Workflow](#workflow)
+      - [Fixing bugs on a prerelease](#fixing-bugs-on-a-prerelease)
   - [Changelog](#changelog)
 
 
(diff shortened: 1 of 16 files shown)


DEVELOPERS.md
```diff
@@ -100,23 +102,55 @@ Fix #issue
...
-
-Run the following command and follow the given instructions.
+Use the following script to release a public version.
 
     npm run publish
 
 
+### Workflow
+
+Everytime a branch is merged to master, an beta version can be released using the `npm run publish:prerelease` command. This command should increment the [semver](https://semver.org/) according to the [commit code](https://www.conventionalcommits.org/en/) and append a `prerelease` code.
```
VictorVelarde: typo: a beta version

IagoLast:
Fixed in #475

PR #484, opened 2018-11-13
IE11 histogram handles

### LINKED ISSUE #479: IE11 - Histogram selection handles not rendering
On IE11, histogram selection works, but it's not rendering the handles.

rjimenezda:
Fix #479 
---

Apparently IE11 doesn't like CSS + SVG very well :(

I checked Builder and we're also not setting width / height / transform on CSS.

I've also removed a typo in the CSS, it might be the reason why the docs are broken


IagoLast:
little sad to me :(

I`d Put a comment poiting we use CSS in JS for IE11 and we are ready to 🚀

PR #492, opened 2018-11-14
Remove height style within widgets

ivanmalagon:
`height: 100%` was causing widgets on FF to not render well, occupying too much space and rendering different from the rest of browsers.

### DIFF
--- a/packages/components/src/components/as-category-widget/as-category-widget.scss
+++ b/packages/components/src/components/as-category-widget/as-category-widget.scss
@@ -10,6 +10,8 @@ as-category-widget {
   --category-bar--color: var(--as--color--complementary);
 
   display: block;
+  min-width: 260px;
+  overflow-y: auto;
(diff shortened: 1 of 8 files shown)


IagoLast:
The [responsive smoke test](http://localhost:5000/packages/components/src/components/as-stacked-bar-widget/responsive) doesn't work anymore.

I think we should research the cause of that bug before removing the height.

rjimenezda:
This is a massive inconsistency between firefox and webkit, and I think firefox is the one that's doing it right. On webkit they do not have 100% height on the sidebar container.

If our histogram / stacked-bar are supposed to take 100% height, we have to surround them with something instead of using them as siblings.

rjimenezda:
If we keep the current implementation (100% height), we should update all examples with such widgets on sidbars, and maybe add a comment somewhere.

ivanmalagon:
Done in https://github.com/CartoDB/airship/pull/502

PR #495, opened 2018-11-14
Redraw histograms / stacked bars when tab becomes active

rjimenezda:
This PR adds an explicit method to the histogram and the stacked bar widget to trigger a redraw externally.

We noticed that if one of these components is placed on an `as-tab` that's not the first one (not initially visible), it won't render correctly. 

We need to be able to notify them somehow that they should re-render. Since we don't have a shared state / event bus / whatever, my approach was to create a helper function that locates `<as-histogram-widget>`'s and `<as-stacked-bar-widget>`'s under a certain element and calls this new method. The `as-tab` and the `as-responsive-content` will do this whenever the active tab changes.

It's not only useful for airship though. If someone uses their own tab system, or a special layout in which the histogram or the stacked bar widget is not initially visible, or has a parent that changes sizes dynamically, they will need to trigger a redraw by hand.

If someone has an idea how to not hardcode the selector I'm all ears :grimacing: 

The only proper alternative I see to using this is to build into airship something like [state-tunnel](https://github.com/ionic-team/stencil-state-tunnel).

### DIFF
--- a/packages/components/src/components.d.ts
+++ b/packages/components/src/components.d.ts
@@ -271,6 +271,7 @@ export namespace Components {
     * Message shown in header when no data is available
     */
     'noDataHeaderMessage': string;
+    'redraw': () => void;
     /**
     * Use this attribute to decide if the widget should be rerendered on window resize. Defaults to true.
     */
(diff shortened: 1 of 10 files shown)


rjimenezda:
**Every** child? So if there's thousands of elements you would el.forceUpdate && el.forceUpdate() for each and one of them?

What if an element in the middle (not a stencil one) has that method for whatever reason, and it should not be called at that moment at all?

The only thing I would maybe accept is duck typing elements on our namespace (as-*), and I think that could still come back to bite us.

IagoLast:
> The only thing I would maybe accept is duck typing elements on our namespace (as-*), and I think that could still come back to bite us.

+1

PR #502, opened 2018-11-15
Refactor footer and sidebar containers

ivanmalagon:
Note: this branch births from branch `firefox-widget-containers`

----

**Context**

While passing the smoke tests, we found critical inconsistencies between browsers layout, specially when combining several widgets in the footer. The footer had a bad experience when filling it up.

The root of all evil in the footer was to try to replicate the fixed and scrollable containers present in sidebars. I believe having fixed containers in the sidebars but no fixed container in the footer is a good trade-off. Footer doesn't need to have that.

Because of this, my proposal in this PR is to simplify containers:
- Footer:
    - No fixed or scrollable container classes: all content in the footer is scrollable.
    - Every piece of content is recommended to be surrounded by `as-box`: it sets a proper flex behaviour and provides with a default padding.
    - A bit taller: widgets got a bit crushed with the given `max-height`.
    - Every `as-box` is responsible for the `overflow` of its content: prior to this, an overflow on a element caused a scroll in the whole footer.
- Sidebar:
    - `container` is still present to provide a fixed container.
    - if the user doesn't need to have a fixed container, it works only with `as-box`.
- Global:
    - `as-box` uses padding instead of margin: this way, we can provide a `--border` modifier that can be used both in sidebar and footer, even if footer has `container` no more.
    - `min-width` of histogram and stacked takes into account the `as-box` padding.

**Next steps**

If you agree with this approach (@IagoLast  and @rjimenezda ) and it passes a whole round of acceptance, we'll need to do the following tasks to close the PR properly:
- [ ] Refactor documentation.
- [x] Adapt specs.
- [ ] Copy the `current` smoke HTML to `cdn` and `npm` smokes.


IagoLast:
`smoke_0` footer on firefox:

[Image from Gyazo](https://gyazo.com/ab2652157815b32949384d0911f343c1)

ivanmalagon:
The tooltip bug is not introduced by this PR. (I commented it in Slack).

ivanmalagon:
Visual tests adapted.

ivanmalagon:
Spotted another bug.

I've added legends to the smokes. When in responsive, the legends appear on top of the map in the map tab. It seems something with the component adding `--visible` when it doesn't have to, because going to the `legends` tab and going back fixes it.

PR #509, opened 2018-11-16
fix(components): Prevent stacked-bar errors when widget is not visible

### LINKED ISSUE #508: StackedBar: Errors thrown when height is 0
**What is happening?**
When the stacked bar widget has no width or height several erros are logged

**What should happen?**
We should control this case to prevent those errors.

IagoLast:
Fix #508 
---

### DIFF
--- a/packages/components/src/components/as-stacked-bar-widget/as-stacked-bar-widget.tsx
+++ b/packages/components/src/components/as-stacked-bar-widget/as-stacked-bar-widget.tsx
@@ -212,6 +212,9 @@ export class StackedBarWidget {
   }
 
   private _drawFigure() {
+    if (this._containerIsNotDrawable) {
+      return;
+    }
     requestAnimationFrame(() => {
       const yAxis = this._drawYAxis();
       this._drawColumns(yAxis);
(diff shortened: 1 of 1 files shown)


packages/components/src/components/as-stacked-bar-widget/as-stacked-bar-widget.tsx
```diff
@@ -212,6 +212,9 @@ export class StackedBarWidget {
   }
 
   private _drawFigure() {
+    if (this._containerIsNotDrawable) {
```
rjimenezda: Uhhh...does this work without parenthesis?

```suggestion
    if (this._containerIsNotDrawable()) {
```

rjimenezda:
:shipit:

packages/components/src/components/as-stacked-bar-widget/as-stacked-bar-widget.tsx
```diff
@@ -256,4 +259,12 @@ export class StackedBarWidget {
       this.el.forceUpdate();
     }
   }
+
+  // If container is not ready or the widget is invisible block drawing phase
+  private _containerIsNotDrawable() {
+    if (!this.container || this.el.clientWidth === 0 || this.el.clientHeight === 0) {
```
ivanmalagon: Do we have `this.el` always available when asking for their properties?

rjimenezda: We should, it's a decorated class property that gets assigned the DOM node of the element.

PR #510, opened 2018-11-16
Refactor footer examples

### LINKED ISSUE #506: Refactor examples on containers
After the last changes in footer and sidebar containers because of the smoke tests, we need to refactor the existing examples.

ivanmalagon:
Fix #506
---


ivanmalagon:
You mean in CSS or where? `as-container` still makes sense in the sidebar to provide fixed and scrollable containers.

PR #526, opened 2018-11-21
Fix histogram for floating values

### LINKED ISSUE #496: Histogram Selection broken for buckets in range [0, 1]
Apparently the histogram selection does not work well when the buckets are in the range [0, 1].

The programmatic API works, but not the mouse events.

It works fine for negative numbers, or mixing negative numbers and positive ones.

rjimenezda:
Fix #496 
---

The real reason why this happens is a Math.round, but it also solves a very annoying case where moving the whole selection could potentially shrink or grow when reaching the start or the end of the histogram.

I've changed it so an internal scale is used that always goes from 0 to # of bins, and is converted back for the event fired.

### DIFF
--- a/packages/components/src/components/as-histogram-widget/utils/draw.service.ts
+++ b/packages/components/src/components/as-histogram-widget/utils/draw.service.ts
@@ -39,6 +39,7 @@ export function updateAxes(
 }
 
 const formatter = format('.2~s');
+const decimalFormatter = format('.2');
 
 export function renderBars(
   data: HistogramData[],
@@ -110,7 +111,15 @@ export function renderXAxis(
     .tickSize(-WIDTH)
     .ticks(3)
     .tickPadding(10)
-    .tickFormat((value) => formatter(realScale.invert(value)));
+    .tickFormat((value) => {
+      const realValue = realScale.invert(value);
+
+      if (realValue > 0 && realValue < 1) {
+        return decimalFormatter(realValue);
+      }
+
+      return formatter(realValue);
+    });
 
   if (container.select('.x-axis').empty()) {
     container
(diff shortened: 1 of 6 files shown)


ivanmalagon:
While it certainly gets better than before, I still find issues.

I've committed a change in the histogram data for the `current` smoke where you can check that there are problems with certain ranges. In particular, with the second and third bucket.

rjimenezda:
I've made some improvements

improvements

rjimenezda:
Because of the '1' position on that GIF, I'm now manually selecting the first, last and middle point as places to print an x-axis tick.

This way it looks better and we have better control of what's being painted and what not. Before, we were asking for 3 ticks but d3-axis decided for us.

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
```diff
@@ -155,7 +181,7 @@ export function renderYAxis(
 }
 
 function _delayFn(_d, i) {
-  return i * 50;
+  return i;
```
ivanmalagon: Does this function do something?

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
```diff
@@ -139,7 +159,13 @@ export function renderYAxis(
     .tickSize(-WIDTH)
     .ticks(5)
     .tickPadding(10)
-    .tickFormat(format('.2~s'));
+    .tickFormat((value) => {
+      if (value > 0 && value < 1) {
```
ivanmalagon: This can be abstracted to a format function and called in the two places it appears.

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
ivanmalagon: I get it. It's for starting animation depending on the number of the bucket, right?

rjimenezda: Exactly, it delays the animation by i millseconds. With i * 50 you get nice results, but if there's too many buckets it's kind of annoying.

With 'i' you don't notice unless there's a lot of buckets.

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
rjimenezda: You're absolutely right.

PR #527, opened 2018-11-22
Make space between bars adapt to available space

### LINKED ISSUE #519: Padding between columns in stacked bars and histograms should be auto or configurable
The space between columns is fixed 
image
but the actual width of the columns get smaller as the number of them grows:
image
If you add enough columns, the result is useless:
image
So, that padding space should adapt the same way the columns width is adapted

rjimenezda:
Fix #519 
---

This PR makes the bars separation a bit smarter. As the code is right now, if you have enough bars and not enough space, the stacked bar might even crash due to negative bar size.

Now, if the projected bar width falls under a certain threshold (3px because why not), the separation between bars is set to 0.

The result is this:
abelqueteparece

Same for the histogram, though it's less likely because the bar separation is 1 instead of 4.

/cc @AbelVm

### DIFF
--- a/packages/components/src/components/as-histogram-widget/utils/draw.service.ts
+++ b/packages/components/src/components/as-histogram-widget/utils/draw.service.ts
@@ -50,14 +50,14 @@ export function renderBars(
   Y_PADDING: number,
   disableAnimation: boolean = false) {
 
-  let BARS_SEPARATION = 1;
+  let barsSeparation = 1;
   const HEIGHT = container.node().getBoundingClientRect().height - Y_PADDING;
   const WIDTH = container.node().getBoundingClientRect().width - X_PADDING;
 
   const barWidth = data.length === 0 ? WIDTH : WIDTH / data.length;
 
-  if (barWidth - BARS_SEPARATION < BAR_WIDTH_THRESHOLD) {
-    BARS_SEPARATION = 0;
+  if (barWidth - barsSeparation < BAR_WIDTH_THRESHOLD) {
+    barsSeparation = 0;
   }
 
   // -- Draw bars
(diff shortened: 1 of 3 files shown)


packages/components/src/components/as-histogram-widget/utils/draw.service.ts
```diff
@@ -43,17 +45,21 @@ export function renderBars(
   yScale: ScaleLinear<number, number>,
   container: Container,
   barsContainer: Container,
-  BARS_SEPARATION: number,
   color: string,
   X_PADDING: number,
   Y_PADDING: number,
   disableAnimation: boolean = false) {
 
+  let BARS_SEPARATION = 1;
```
ivanmalagon: If it's not a constant anymore, let's switch it to lowercase.
```suggestion
  let barsSeparation = 1;
```

There are more places in the file with reference to it.

packages/components/src/components/as-stacked-bar-widget/as-stacked-bar-widget.tsx
```diff
@@ -226,9 +228,16 @@ export class StackedBarWidget {
       return;
     }
     const Y_AXIS_LABEL_WIDTH = 25; // We draw on the right of the yAxis labels
-    const COLUMN_MARGIN = 4;
-    const WIDTH = yAxisElement.getBoundingClientRect().width - Y_AXIS_LABEL_WIDTH - COLUMN_MARGIN;
-    const COLUMN_WIDTH = (WIDTH / this.data.length) - COLUMN_MARGIN;
+    let COLUMN_MARGIN = 4;
```
ivanmalagon: Same as above. This is no longer a constant.

PR #539, opened 2018-11-29
Time series widget

### LINKED ISSUE #534: Time Series Widget
We need to create a widget similiar to the histogram that supports several new features:

- Having dates or epochs as start / end bucket ranges.
- Animation status / seeking.
- Different formatting for the X-axis (Years, Years+Month...etc)

The animation part should be entirely optional, and on a very basic case, it should pretty much look like a regular histogram.

Related ticket #457

rjimenezda:
Fix #534 
---

This PR adds a new component that composes a Histogram and adds time data and animation status support.

The time support simply means that the x-axis is properly labeled with time data. The histogram now supports an x-axis-formatter function prop.

We probably should have the same prop on the Time Series so that users can provide the proper formatter based on their data aggregation (full date, months, quarters...etc). I think this would be better than adopting a string pattern that other libs like d3-time-format or moment support.

The animation part is completely optional, and if the prop is not there, it should look like a regular histogram.

If the animation property is there: 

- A Play / Pause button is added, which simply fires an event. We could consider removing this altogether and letting users compose their own version.
- A scrubber will be rendered, based on the progress ([0, 100]) prop.
- The scrubber will render inside the selection if there's any.
- Clicking on the histogram will trigger a seek event, to notify the animation to move forward.
- There's a little video-player-like hint on the scrubber.

I've tried to make it as compatible as possible with the brush.

timeseries

### DIFF
--- a/packages/components/src/components/as-histogram-widget/as-histogram-widget.tsx
+++ b/packages/components/src/components/as-histogram-widget/as-histogram-widget.tsx
@@ -171,6 +171,7 @@ export class HistogramWidget {
   @Prop() public axisFormatter: (value: number | Date) => string;
 
   public selection: number[] = null;
+  public _lastEmittedSelection: number[] = null;
 
   @Element() private el: HTMLElement;
 
@@ -361,7 +362,7 @@ export class HistogramWidget {
           this._onBrushEnd.bind(this),
           CUSTOM_HANDLE_WIDTH,
           CUSTOM_HANDLE_HEIGHT,
-          X_PADDING,
+          X_PADDING + (this.yLabel ? LABEL_PADDING : 0),
           Y_PADDING
         );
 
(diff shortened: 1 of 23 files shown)


packages/components/src/components/as-histogram-widget/as-histogram-widget.tsx
```diff
@@ -509,8 +468,8 @@ export class HistogramWidget {
   }
 
   private _hideCustomHandlers() {
```
ivanmalagon:
```suggestion
  private _hideCustomHandles() {
```

packages/components/src/components/as-histogram-widget/utils/brush.service.ts
```diff
@@ -0,0 +1,95 @@
...
+      .attr('stroke-width', 4)
+      .style('opacity', 0)
+      .attr('pointer-events', 'none');
+  }
+
+  bottomLine
+    .attr('y1', height - Y_PADDING)
+    .attr('y2', height - Y_PADDING);
+
+  let customHandlers = brushArea.selectAll<SVGGElement, { type: string }>('.handle--wrapper');
```
ivanmalagon: Handlers or handles?

rjimenezda: It's handles, I noticed this but totally forgot to change it.

rjimenezda:
Closing this PR in favor of the Bindings one which contains everything

PR #544, opened 2018-12-12
As vl bindings

### LINKED ISSUE #536: Basic VL Integrations
We should provide easy to use integrations for our map libraries. 

The outcome of this issue should be a new package on airship that contains bindings to common use cases of CARTO VL + Airship

- [ ] Decide on a name, **very important**.
- [ ] Minimal examples.

We should aim for basic examples that don't take internal state into account (i.e: filtering)

- Histogram
- Category widget
- Layer selection :question:

### LINKED ISSUE #537: Animation VL integration
This issue assumes #534 and #536 are finished.

The outcome of this issue should be new bindings that let you pass a VL animation and properly do all the piping so you can control it through an airship time series:

- [ ] Play / pause
- [ ] Update the animation progress on the widget.
- [ ] Seeking

Extra credit:

- [ ] Change color of buckets that have already been played

### LINKED ISSUE #538: Filtering VL integration
Filtering through a widget and re-setting the state to the filtered data can be [very tricky](https://github.com/CartoDB/cartodb/tree/master/lib/assets/javascripts/deep-insights)

We should try to provide a standardized way of filtering VL visualizations with Airship widgets.

- [ ] Histogram Widget
- [ ] Time Series Widget
- [ ] Category Widget

We could tackle this with a simple approach first (no re-setting the filtered data), and then also providing a more advanced one.

We might have to impose some limitations in order to keep this feasible. For instance, what happens if you move the map while you've filtered, or if the same visualization is filtered by several widgets, etc etc

rjimenezda:
Fix #536 
Fix #537
Fix #538
---

This started like a little function that did all the animation stuff for you and turned into the monster that lies before you.

This PR enables some bindings / shortcuts to:

- Bind columns to widgets easily
- Have multi-widget filtering (just like on builder). This is experimental and not a good idea on big datasets, though.
- Bind a VL animation variable to a time series effortlessly.

Please check DEVELOPERS.md for some of my thoughts on this, and an schema of how it works internally.

There are examples using this on time-series-widget and histogram-widget.

The multiple widget selection works like on builder (for the most part) but it's incredibly expensive. There's a proposal on DEVELOPERS.md for an alternative that should be cheaper, but I think we should consider adding some extra stuff to VL.

Remaining important things we have to do (not necessarily on this PR)

- [ ] Fix some rough edges on animating a range
- [ ] Setup minimal testing / linting
- [ ] Examples
- [ ] Build for CDN and for NPM (it's typescript, so can't upload the src)
- [ ] Publish this package on NPM (CDN should work with minimal effort)

I would suggest making this an alpha / beta, so documenting can be drafty and can afford to break the API for a little bit. I've taken some decisions like asking for an actual Viz instead of a string, maybe that's stupid and I didn't notice, and so on.


rjimenezda:
For the build / distribution:

Right now I'm compiling TS and bundling with webpack into a single file. This is great for the CDN, but kind of annoying for NPM.

There's not much to tree shake right now, though, so I wonder if it's worth it uploading something different to NPM. We could upload the bundle and also the TS source code / definitions.

rjimenezda:
`time-series` branch (this branch's original parent) has a few commits (
721b991...ffc7cda) that this one is lacking. We should merge it into this one and remove the other PR, an hope for the best :see_no_evil:

PR #550, opened 2018-12-30
Fix introduction guide

VictorVelarde:
Fix https://github.com/CartoDB/airship/issues/551
---
Update mapbox version used in the guide, removing our old fork, now that we support the official mapbox version (from v0.52).


VictorVelarde:
We have been talking about releasing a v1.1 soon, so maybe it's better to keep this open 'till then, and then update all thee references properly

VictorVelarde:
Ready for VL v1.2

Issue #552, opened 2018-12-30
Missing references in Getting started guide


VictorVelarde:
**What is happening?**
Getting started guide is missing mapbox and CARTO VL libraries, so if anybody uses the 'basic-template' as-it-is he gets an error (mapbox is not defined)

**What should happen?**
Copy-paste code should work

**Steps to Reproduce**
See https://carto.com/developers/airship/guides/getting-started/#basic-template and try to create an example as it is.

PR #553, opened 2018-12-30
Include missing mapbox & CARTO VL references in basic template

VictorVelarde:
Fix https://github.com/CartoDB/airship/issues/552


VictorVelarde:
We have been talking about releasing a v1.1 soon, so maybe it's better to keep this open 'till then, and then update all thee references properly

VictorVelarde:
Updated to v1.2.0 (to be released soon...)

rjimenezda:
Pointed this to `v2`

PR #561, opened 2019-01-18
Histogram / Time Series multiline tooltip support

rjimenezda:
Branch chart:

master -> v2 -> histogram-clear -> histogram-tooltip

image

This PR adds support for the tooltipFormatter prop to return an array of strings, which renders them each on a line.

Added an example for this and updated the documentation.

I've also added some extra @Methods. One is kind of annoying, because it's a proxy to a prop, but it's a prop that's set on the time-series and we need to be able to acces from outside.

Also, the time series was lacking all the histogram methods, so you couldn't really programmatically get the selection, clear it etc. I've added the proxy methods


rjimenezda:
Tests failing for completely unrelated component, so :man_shrugging:

PR #562, opened 2019-01-21
Set min value or max value when mouse is past edge bounds

jesusbotella:
This PR sets min or max value when mouse is past range slider bounds.

There were some cases when sliding mouse so fast to the left didn't fire some `mousemove` events that are necessary to set slider thumb to the left most position. It should happen with the right one as well.

### DIFF
--- a/packages/components/src/components/as-range-slider/thumb/as-range-slider-thumb.tsx
+++ b/packages/components/src/components/as-range-slider/thumb/as-range-slider-thumb.tsx
@@ -122,12 +122,16 @@ export class RangeSliderThumb {
     const eventX = changedTouches ? changedTouches[0].pageX : (event as MouseEvent).pageX;
     const barPercentage = (eventX - this.railBoundingClientRect.left) * 100 / this.railElement.offsetWidth;
 
-    if (barPercentage < 0) {
+    if (barPercentage < 0 && this.percentage > 0) {
       return this.thumbMove.emit(0);
     }
 
-    if (barPercentage > 100) {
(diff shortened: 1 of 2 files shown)


packages/components/src/components/as-range-slider/thumb/as-range-slider-thumb.tsx
```diff
@@ -122,8 +122,12 @@ export class RangeSliderThumb {
     const eventX = changedTouches ? changedTouches[0].pageX : (event as MouseEvent).pageX;
     const barPercentage = (eventX - this.railBoundingClientRect.left) * 100 / this.railElement.offsetWidth;
 
-    if (barPercentage < 0 || barPercentage > 100) {
-      return;
+    if (barPercentage < 0) {
+      return this.thumbMove.emit(0);
```
rjimenezda: My only minor concern with this is that if you'll get a lot of 'public' events thrown if you keep dragging after reaching the beginning or the end. I'd try to throttle it somehow on the parent component.

However, it's totally consistent with the current behavior for other steps. When dragging between two steps it fires n events, and not just one when the value changes, so I wouldn't worry too much about it. :man_shrugging:

jesusbotella: Yes, I tried to limit it by checking that percentage was different to 0 (`barPercentage < 0 && this.percentage !== 0`). But as percentage might not be exactly 0, but 0.0004 we need to use a "fuzzy" comparison to achieve that behaviour :(

PR #563, opened 2019-01-23
Select bucket when clicking it (only for the histogram)

rjimenezda:
master -> v2 -> histogram-clear -> histogram-tooltip -> click-bucket-select

This PR adds the ability to select a bucket by clicking on it.

This is only available to the histogram, not the time series. This also happens on builder because of the same problem, we need to be able to click on the time series to seek the animation to a certain percentage, so we can't have both events.


rjimenezda:
Tests failing for unrelated component so :man_shrugging:

PR #567, opened 2019-02-01
Histogram Axis options

rjimenezda:
This PR adds extra options to the histogram to control details related to the axis. To be more precise:

- Lets you define a custom range (Y-Axis)
  - This is very useful to appreciate when filtering / showing a subset of data
- Proxies a few options of d3-axis:
  - Which ticks will be shown (axis.tickValues)
  - How many will be shown (axis.ticks)
  - A custom formatter for the axis (axis.format)
  - Padding (axis.padding)

Warning: tickValues has precedence over ticks.

I've also added a prop to disable the animations, which could be tedious for histograms with a lot of buckets (animation delay is bucket index).

Custom range example:
image

Axis options examples:
image

Missing:
- [x] Proxy from time series
- [x] Documentation for dev center


rjimenezda:
I've also fixed some weird behaviour with negative values. Now they're rendered above or below the axis.

I know it makes no sense for a histogram, but since there's no plan for a BarChart, we might as well 'fail' gracely. It won't be an official feature, though, so no docs for that.

PR #573, opened 2019-03-11
Global histogram support on VL Bridge

rjimenezda:
This adds support for the global histogram on the VL bridge for numerical histograms.

There's a new `totals` argument you can provide.

In order for this to work properly, you either have to specify the buckets yourself, or the viewportHistogram will wait for the globalHistogram to be ready, and get the bucket list from there.

This is because in order to represent them both on the same x-axis, the buckets must be the exact same.


rjimenezda:
I'm including a VL compilation for the examples until the new version is pushed to the CDN <3

PR #574, opened 2019-03-11
Range slider filter for VL Bridge

rjimenezda:
This PR adds a `global` range slider filter to the Bridge. This lets you easily filter a range for number columns using an as-range-slider widget. Underneath it uses globalMax and globalMin, and a `$x > y and $x < z` filter

I mentioned this in passing to @makella @elenatorro and @jorgesancha while demoing the bridge and thought, hey, might as well add it since it's a low hanging fruit.


makella:
:boom:

PR #576, opened 2019-03-20
Screencast suggestions

rjimenezda:
The following changes have been added to the VL Bridge. I'm going to ping the `suggesters`, so if you got notified look for your name below:

- Implicit animation (@VictorVelarde @elenatorro ), now there's no requirement on what the viz contains.
- API Changes (options are now actually optional :joy: ) @jesusbotella 
- Category Widget supports a button option to delegate the filtering (@mbforr )
- Documented 'widget side effects', which is the Airship Properties the filter are changing.
- Histograms can take expressions as arguments (@elenatorro)
- Categorical Histograms support global histograms now as well
- The auto color features are still there, but I've removed the docs for now because we can't make it work nicely for numerical histograms (@makella )


packages/bridge/src/vl/time-series/README.md
```diff
@@ -10,57 +10,58 @@ Internally, a histogram is created to display the bars, and also all the necessa
...
 ### Usage
 
-There is one caveat for this case. In order to have more freedom on the animation itself, we do not create it internally. However, this means that we expect the animation on a particular variable. 
+There are two ways of using this:
 
-In order for this filter to work, please define your animation as a variable called `@animation`:
+- Creating an animation expression on the viz, stored on a variable
+- Automatically creating an animation where you can control certain parameters
+
+If you create the animation expression, you can call it `@animation` or if you need to, name it as you wish and specify such name on the options object of this.
```
VictorVelarde: ✨

packages/bridge/src/vl/time-series/README.md
```diff
@@ -70,6 +71,9 @@ BucketRanges = [number, number]
 `readOnly` is a boolean to specify whether this widget should filter or not.
 `nBuckets` is a number to specify the number of buckets the histogram should have
 `bucketRanges` is an array of pairs of numbers, which lets you specify the bucket ranges by hand
+`duration` how long should the animation be, defaults to 30
```
VictorVelarde: It would be nice to add the units, which are 'seconds'

packages/bridge/src/vl/time-series/TimeSeries.ts
```diff
@@ -116,6 +143,23 @@ export class TimeSeries {
       this._animation.pause();
     });
   }
+
+  private _createAnimation() {
+    const s = this._carto.expressions;
+
+    return s.animation(
```
VictorVelarde: I see your point here on creating something more easily from the string representation

packages/bridge/src/vl/time-series/README.md
rjimenezda: :+1: Done in another branch, I've also changed it to 10, which is VL's default

packages/bridge/src/vl/time-series/TimeSeries.ts
rjimenezda: Yep, VL String API is :boom:

PR #577, opened 2019-03-25
Fix icons pollution

rjimenezda:
This PR makes it mandatory to add an `as-icon` class to i elements that want to use icons.

As it was before, it was tricky to use `<i>` elements for anything else, as all would get the airship iconfont assigned to them.


rjimenezda:
I did the CSS update manually because I don't want to go to icomoon again.

I've also removed the `Hamburguer` icon, in favor of the correct Hamburger spelling

docs/guides/06-icons.md
```diff
@@ -25,17 +25,17 @@ First, you need to include our `icons.css` file in the head of your app.
   <link rel="stylesheet" href="https://libs.cartocdn.com/airship-icons/%AS-VERSION%/icons.css">
 </head>
 
-Then you can use the icons inside an `i` tag by adding the `as-icon-` class followed by the icon name you want to use.
+Then you can use the icons inside an `i` tag by adding the `as-icon` class, and also the `as-icon as-icon-` class followed by the icon name you want to use.
```
VictorVelarde: Should we remove the 'as-icon' class in the second occurrence?

rjimenezda: Yep

PR #595, opened 2019-04-04
Add CORS options to serve

### LINKED ISSUE #593: Serving Airship locally should have CORS enabled
We're doing this using `serve`, and we have to write a small serve.json config file to send CORS headers.

This is required for other projects and is overall helpful.

rjimenezda:
Fix #593 
---


rjimenezda:
Styles are broken because circleci is not installing node-sass

PR #605, opened 2019-05-24
Use getLegendData for opacity

elenatorro:
Fix https://github.com/CartoDB/carto-vl/issues/1353
Screenshot 2019-05-24 at 17 42 49

---

Use getLegendData for opacity, which is used in CARTOFrames helpers


Jesus89:
Vamos!

Issue #604, opened 2019-05-23
Overwrite label for CARTO_VL_OTHERS


Jesus89:
**Problem**
The legend shows CARTO_VL_OTHERS.

**Proposed solution**
It should show "Others" by default and it would be nice to have the ability to pass this name by parameter.

PR #606, opened 2019-05-24
Customize others label

elenatorro:
Fix https://github.com/CartoDB/airship/issues/604
---

Related CARTOframes PR https://github.com/CartoDB/cartoframes/pull/699

### DIFF
--- a/packages/bridge/src/vl/legends/index.ts
+++ b/packages/bridge/src/vl/legends/index.ts
@@ -136,6 +136,7 @@ interface LegendOptions {
   onLoad?: () => void;
   // Should the legend repaint after layer updates or just initially
   dynamic?: boolean;
+  config?: (othersLabel?: string, samples?: number) => object;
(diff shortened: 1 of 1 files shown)


packages/bridge/src/vl/legends/index.ts
```diff
@@ -136,6 +136,7 @@ interface LegendOptions {
   onLoad?: () => void;
   // Should the legend repaint after layer updates or just initially
   dynamic?: boolean;
+  config?: (othersLabel?: string, samples?: number) => object;
```
rjimenezda: I think the typing should be:

```suggestion
  config?: { othersLabel?: string, samples?: number };
```

As it is, it means a function that takes those 2 arguments and returns an object, doesn't it?

elenatorro: ups, yes, you're right!

PR #614, opened 2019-06-04
Line size legend component

### LINKED ISSUE #612: Line size legend component

### LINKED ISSUE #615: Multi-layer legend font sizes are different
**What is happening?**
In CF integration, with multi-layer helper methods, I am noticing that there is a difference in font size between the two layers legend labels.

in this case the range values (`color_bins_layer`):
Screen Shot 2019-06-11 at 10 15 27 AM

in this case bins + category:
Screen Shot 2019-06-11 at 11 21 34 AM

In both cases, it looks like the top legend's labels are larger than the bottom layer's legend.

**What should happen?**
Both should be equal.

**Steps to Reproduce**
Create a multi-layer map with helper methods in CF. 

cc @rjimenezda

rjimenezda:
Fix #612 
Fix #615 
---

image

Created the base `<as-legend-line-size>` component and added the following aliases:

- `<as-legend-size-bins-line>`
- `<as-legend-size-continuous-line>`

And now, `<as-legend-size-bins>` will render it if the type is 'line', and so will `<as-legend-size-continuous>`

~Unrelated tests are failing despite not having changed anything :thinking:~ A rerun fixed it


packages/components/src/components.d.ts
```diff
@@ -1497,10 +1531,13 @@ declare global {
     'as-legend-color-continuous-point': Components.AsLegendColorContinuousPointAttributes;
     'as-legend-color-continuous-polygon': Components.AsLegendColorContinuousPolygonAttributes;
     'as-legend-color-continuous': Components.AsLegendColorContinuousAttributes;
+    'as-legend-size-bins-line': Components.AsLegendSizeBinsLineAttributes;
     'as-legend-size-bins-point': Components.AsLegendSizeBinsPointAttributes;
     'as-legend-size-bins': Components.AsLegendSizeBinsAttributes;
+    'as-legend-size-continuous-line': Components.AsLegendSizeContinuousLineAttributes;
     'as-legend-size-continuous-point': Components.AsLegendSizeContinuousPointAttributes;
     'as-legend-size-continuous': Components.AsLegendSizeContinuousAttributes;
+    'as-legend-size-line': Components.AsLegendSizeLineAttributes;
```
Jesus89: I think we don't need this component

rjimenezda: That's the real component, the other two are just proxying it right now. I think this is more semantic than naming it 'as-legend-size-continuous' or 'as-legend-size-bins' and then using it on the other one.

If we end up having differences between the two then we can split it into two different ones.

rjimenezda:
I uploaded the continuous one.

image

This component is slightly different from the rest because it's done with SVG. so it's less flexible size-wise. You can specify a size property (300px by default), and even a factor that is used to add a bit of inner margin (0.1 by default)

When it's used horizontally (not the case for cartoframes) there's some extra logic to reduce the required space for the SVG.

On the vertical rendering, this doesn't happen. Partly because we can use the space to render long numbers without worrying too much.

rjimenezda:
image

Now there's also as-legend-size-category which is a metacomponent for:

- `as-legend-size-bins-point`, which now supports `marker` as well.
- The new `as-legend-size-category-line`

packages/components/src/components/legends/as-legend-size-continuous-line/as-legend-size-continuous-line.tsx
```diff
@@ -0,0 +1,210 @@
...
+    const sortedData = this.data.slice().sort(
+      (first, second) => second.width - first.width
+    );
+
+    const max = sortedData[0].width;
+    const half = max / 2;
+
+    const X_POS = Math.max(max + MIN_LINE_SIZE, this.width);
+
+    let X_OFF = (this.width - max) / 2;
```
elenatorro: `X_OFF` doesn't need to be modified, it should be a constant. 

```js
const X_OFF = this.orientation === 'horizontal' || this.width === null
    ? 0
    : (this.width - max) / 2;
```

packages/components/src/components/legends/as-legend-size-continuous-line/as-legend-size-continuous-line.tsx
```diff
@@ -0,0 +1,210 @@
...
+  public componentWillUpdate() {
+    this.parseSize();
+  }
+
+  public componentWillLoad() {
+    this.parseSize();
+  }
+
+  public render() {
+    const sortedData = this.data.slice().sort(
```
elenatorro: We're using this method several times, I think it could be a getter. I know it's being used in more components, but maybe for this one we could use a getter to avoid duplicity.

packages/components/src/components/legends/as-legend-size-continuous-line/as-legend-size-continuous-line.tsx
```diff
@@ -0,0 +1,210 @@
...
+      (first, second) => second.width - first.width
+    );
+
+    const max = sortedData[0].width;
```
elenatorro: I would always use capital case for constant numbers: MAX, HALF, etc. 

Also, I'd use `X_OFFSET` and `Y_OFFSET` (instead of yOffset). If it's common to use `yOffset`, I'd use it then for both X and Y offsets.

packages/components/src/components/legends/as-legend-size-continuous-line/as-legend-size-continuous-line.tsx
```diff
@@ -0,0 +1,210 @@
...
+      if (this.orientation === 'horizontal') {
+        top.reverse();
+        bottom.reverse();
+      }
+
+      top[0] += X_OFF;
+      bottom[0] += X_OFF;
+
+      // Insert always at current index + 1 (to account for `M0 ${this.size}`)
+      realPath.splice(i, 0, `L${top.join(' ')}`);
```
elenatorro: Before checking if there's a label, I think we can also have the line 100. With capital case constants, it'd be something like:

```js
        const FACTOR = HALF * (d.width / MAX);
        const Y_RATIO = i / (sortedData.length - 1);
        const TOP_X = HALF + FACTOR;
        const BOTTOM_X = HALF - FACTOR;
        const TOP_Y = BOTTOM_Y = this.size - (this.size * Y_RATIO);

        const top = this.orientation === 'vertical'
            ? [TOP_X + X_OFFSET, TOP_Y]
            : [TOP_Y + X_OFFSET, TOP_X];

        const bottom = this.orientation === 'vertical'
            ? [BOTTOM_X + X_OFFSET, BOTTOM_Y]
            : [BOTTOM_Y + X_OFFSET, BOTTOM_X];

        path.splice(i, 0, `L${top.join(' ')}`);
        path.splice(path.length - i, 0, `L${bottom.join(' ')}`);

        // label logic
```

rjimenezda:
I've made elena's suggestions.

I'm going to keep this open to merge the specs branch into this one later.

PR #616, opened 2019-06-12
Legend specs

### LINKED ISSUE #613: Add tests for legend components
Right now there's several legend components that have no testing whatsoever.

We should at least add snapshot tests to all the significant ones, and perhaps to some of the metacomponents.

If any has some particular behaviour (like the category one, width propagation) we should probably add e2e tests there as well.

rjimenezda:
Fix #613 
---


rjimenezda:
Circle CI is acting up, tests are taking a loooong time, and some are failing seemingly randomly :thinking:

rjimenezda:
Ok, so I've made it so jest runs the tests one after the other instead of using workers, which I guess was hitting some limits in CircleCI

It might take longer in the long run, but it doesn't randomly fail. Also, if you use the :watch command locally, they don't run sequentially.

PR #623, opened 2019-06-13
Fix size point distorted

### LINKED ISSUE #617: circle distorted in size bins point legend
**What is happening?**
It looks like some of the circles on the size bins point legend are distorted.. for example, the last bin here:
Screen Shot 2019-06-12 at 2 25 23 PM

elenatorro:
Fix #617
---

Before:
Screenshot 2019-06-13 at 14 31 35

After:
Screenshot 2019-06-13 at 14 21 59


packages/components/src/components/legends/as-legend-size-bins-point/as-legend-size-bins-point.tsx
```diff
@@ -44,26 +44,31 @@ export class LegendSizeBinsPoint {
...
 
     if (this.orientation === 'horizontal') {
-      wrapperStyle.height = `${this.maxSize}px`;
+      wrapperStyle.height = MAX_SIZE_PX;
     } else if (this.orientation === 'vertical') {
-      wrapperStyle.width = `${this.maxSize}px`;
+      wrapperStyle.width = MAX_SIZE_PX;
     }
 
+    const STYLE_SIZE_PX = SIZE > MAX_SIZE ? MAX_SIZE_PX : SIZE_PX;
```
rjimenezda: So, if you do this you're making the point smaller than it needs to be. The 2 pixels are being added to account for the border and box-sizing: border-box.

I think what we should do is make the wrapper Math.max(SIZE, MAX_SIZE).

elenatorro: I see! You're right, let's try that

PR #629, opened 2019-06-24
Feature/626 upgrade stencil

### LINKED ISSUE #626: Upgrade to stencil 1.0
We're two versions behind, and they've released the first 'stable' version, so we should upgrade for our next minor.

We will have to make all public methods async, but that's already the case so it should not be a breaking change, it's just that it might not compile if we don't.

We should take a look at how their doc generation works, the components documentation is too disconnected right now and we're not taking advantage of typescript _at all_

rjimenezda:
Fix #626 
---

This is not being too bad for now, no apparent breaking changes. Some notes:

- I had to add a bunch of async to public methods. They were already async but now they're documented
- A lot of snapshots are broken because of a few characters :roll_eyes:
- E2E tests are super slow now, might move a bunch of them to unit tests at least for snapshots.
- The `module` of the package is changed to another path.
~- One spec (range slider) commented for now because `draggable` is a reserved attribute, and the spec runner fails those tests. I opened a PR to stencil to fix that, it's been merged but we'll have to wait for the patch.~ 1.1.0 included this change
- I've added an extra prop (`isDraggable`) and marked `draggable` for deprecation for the next major to keep it clean. We should somehow track this stuff so we don't forget. On the next major we should also unify a bunch of stuff like showClear / showClearbutton etc.
- Stencil has changed the way to include the files in a page, but they still support the old one with a warning. I'll update the docs to use the new one, and note the old one works for now.
- Infowindow was completely broken because of a hacky implementation, fixed
- There are _a lot_ more files now on the build output.


rjimenezda:
1.1.0 fixed the general slowness on the tests, we're back to ~30 secs on my machine which I think is decent.

We could redo tests to not use puppeteer but I'd rather not touch too much just in case.

rjimenezda:
Never mind, now they're randomly failing possibly because of a lack of resources on Circle, so we might have to refactor a few into using the puppeteerless approach

PR #634, opened 2019-06-28
Add Animation Controls widget

elenatorro:
Fix https://github.com/CartoDB/cartoframes/issues/812
---

Add a simple Animation Controls widget and its VL Bridge connector

Screenshot 2019-06-28 at 14 31 21


packages/components/src/components/as-range-slider/slider/as-range-slider.tsx
```diff
@@ -141,28 +157,31 @@ export class RangeSlider {
...
+        onThumbChangeStart={() => this._emitChangeIn(this.changeStart)}
+        onThumbChangeEnd={() => this._emitChangeIn(this.changeEnd)}>
+      </as-range-slider-thumb>;
+    }
   }
 
   private _renderRangeBar() {
     const [firstThumbPercentage, lastThumbPercentage] = this._getCurrentThumbPercentages();
-    const draggable = (this.isDraggable || this.draggable) && this.range !== undefined;
+    const isDraggable = (this.isDraggable || this.isDraggable) && this.range !== undefined;
```
rjimenezda:
```suggestion
    const isDraggable = (this.isDraggable || this.draggable) && this.range !== undefined;
```

packages/components/src/components/as-range-slider/slider/as-range-slider.tsx
```diff
@@ -54,25 +54,41 @@ export class RangeSlider {
   /**
    * Disables component if truthy
    *
-   * @type {number}
+   * @type {boolean}
```
rjimenezda: Good catch :eyes:

packages/components/src/components/as-range-slider/slider/as-range-slider.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { RangeSlider } from './as-range-slider';
 
 /**
- * Commented until we can deprecate the prop "draggable" because stencil is throwing an error when
+ * Commented until we can deprecate the prop "isDraggable" because stencil is throwing an error when
```
rjimenezda:
```suggestion
 * Commented until we can deprecate the prop "draggable" because stencil is throwing an error when
```

packages/components/src/components/as-range-slider/track/as-range-slider-bar.scss
```diff
@@ -17,7 +17,7 @@ as-range-slider-bar {
     background-color: var(--as--range-slider__range-bar--background-color--disabled);
   }
 
-  .as-range-slider__range-bar--draggable {
+  .as-range-slider__range-bar--isDraggable {
```
rjimenezda: You know, I'm totally OK with this but I wonder if someone would consider changing class names a breaking change :thinking:

packages/components/src/components/as-range-slider/slider/as-range-slider.tsx
```diff
@@ -54,25 +54,41 @@ export class RangeSlider {
...
    * @memberof RangeSlider
    */
   @Prop() public disabled: boolean = false;
 
+  // /**
```
rjimenezda: Uncomment and rename to draggable

examples/bridge/animation-controls/simple.html
```diff
@@ -0,0 +1,91 @@
...
+      <main class="as-main">
+        <div class="as-map-area">
+          <div id="map"></div>
+        </div>
+        <div class="as-map-footer">
+          <div class="as-box">
+            <as-animation-controls-widget
+              animated
+              heading='Animation'
+              description='Try seeking, pausing or selecting a range'
```
rjimenezda: Selecting a range doesn't work on this one so we should remove it

```suggestion
              description='Try seeking and pausing'
```

packages/components/src/components/as-animation-controls-widget/as-animation-controls-widget.tsx
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * Use this attribute to put the widget in "loading mode".
+   * When loading mode is active, a spinner will be shown and the data will be hidden.
+   */
+  @Prop() public isLoading: boolean = false;
+
+  @Prop() public showThumb: boolean = true;
+  @Prop() public showThumbCaption: boolean = false;
+  @Prop() public playing: boolean = false;
+  @Prop() public progressValue: number = 0;
```
rjimenezda: I think this prop is unused

examples/bridge/animation-controls/simple.html
```diff
@@ -0,0 +1,91 @@
...
+              time-format='%Q'
```
rjimenezda: Also this is not a prop in this case

packages/components/src/components/as-animation-controls-widget/as-animation-controls-widget.scss
```diff
@@ -0,0 +1,32 @@
...
+
+as-animation-controls-widget {
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  min-width: $widget-min-width;
+  height: 100%;
+  max-height: 100%;
+  overflow: auto;
+  background: var(--as--histogram-widget--background-color);
```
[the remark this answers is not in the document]
elenatorro: You're right, just added a variable for this widget's background

VictorVelarde:
Have we considered adding the timestep visually (over or under the range slicer)? I think we should. With that piece, I think we cover a much higher number of use cases.

Some ideas we could use:
- a boolean property to display it: 'none', 'top', 'bottom'
- a function receiving a date, to control the formatting

What do you think?

rjimenezda:
I think the idea was to keep it simple. Having it on the scrubber makes it tricky to format (overflow on the sides). Also I don't know if the range sliders supports having it elsewhere.

We already have some things to format dates due to the as-time-series-widget (it receives a format as per d3-time-format).

elenatorro:
I like the idea, actually there're other things it'd be great to have, like showing the current value of the animation. As Román says, we idea is to keep it simple now and create all the widgets, see how they work together, and then improve them. It'd also be great if the design team could be involved, but I'd do that in the next iteration.

VictorVelarde:
@elenatorro @rjimenezda I was talking about 'current value of the animation' indeed 😄 , but it's clear that I didn't explain myself correctly.

I aggree with not adding it now, in this first version, but I think that displaying the date will be eventually required. Let's wait then.

andy-esch:
I just wanted to chime in that I agree with @VictorVelarde and think it's important for the next iteration of this widget to have the current timestamp (or number if animating on a numerical field). This helps the person viewing the map and interacting with it to understand the _when_ and where of the events in the data. I also love @rjimenezda's idea of having a formatting string so the user can control how the timestamp (or number) is shown on the map.

PR #635, opened 2019-07-03
Change animation controls API

elenatorro:
Improve animations-controls-widget API
---

- [x] Use a variable from the viz style
- [x] Show progress value
- [x] Add examples
- [x] Add documentation in README.md
- [x] Add bridge documentation in README.md


packages/bridge/src/vl/animation-controls/AnimationControls.ts
```diff
@@ -97,13 +98,33 @@ export class AnimationControls {
...
   }
 
+  private _formatProgressValue() {
+    const progressValue = this._animation.getProgressValue();
+
+    if (progressValue instanceof Date) {
+      return this._formatCb ? this._formatCb(progressValue) : progressValue.toISOString();
+    }
+
+    if (this._isVLTimeZoneDate(progressValue)) {
```
jesusbotella: I think we should extract this functions as they're not specific to this class and may be reused.

packages/bridge/src/vl/animation-controls/AnimationControls.ts
```diff
@@ -1,7 +1,8 @@
-import { VL_BINARY_EXPRESSION_TYPES, VLAnimation, VLViz } from '../../types';
+import { VL_BINARY_EXPRESSION_TYPES, VLAnimation, VLTimeZoneDate, VLViz } from '../../types';
 import { select } from '../../util/Utils';
 
 export class AnimationControls {
+  protected _formatCb: (value) => void;
```
jesusbotella: Being specific, the value should be `Date | number` and should return a `string`, right?

And why is this protected? Is this class being inherited somewhere?

examples/bridge/animation-controls/date-format.html
```diff
@@ -0,0 +1,92 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <title>Airship bridge</title>
+
+  <script src="https://cdn.jsdelivr.net/npm/lodash@4.17.11/lodash.min.js"></script>
```
jesusbotella: Lodash? 😁

examples/bridge/animation-controls/date-format.html
```diff
@@ -0,0 +1,92 @@
...
+    const timeWidget = document.querySelector('as-animation-controls-widget');
+
+    const bridge = new AsBridge.VLBridge({
+      carto: carto,
+      map: map,
+      layer: layer,
+      source: source
+    });
+
+    bridge.animationControls('as-animation-controls-widget', 'hour', { variableName: 'myVariable' });
```
jesusbotella: We're not setting the animation expression to any variable named `myVariable`, right?

examples/examples.json
```diff
@@ -205,6 +205,22 @@
...
+          "title": "animation-controls-widget simple",
+          "file": "components/as-animation-controls-widget/simple.html"
+        },
+        {
+          "title": "animation-controls-widget display dates",
+          "file": "components/as-animation-controls-widget/date-format.html"
+        },
+        {
+          "title": "animation-controls-widget animate a property",
+          "file": "components/as-animation-controls-widget/property.html"
```
jesusbotella: we're missing this example?

packages/bridge/src/vl/animation-controls/README.md
```diff
@@ -0,0 +1,86 @@
...
+  propertyName?: string;
+}
+```
+
+`duration` how long should the animation be in seconds, defaults to 10
+`fade` an array of two durations for fading in and out, check the VL Animation [documentation](https://carto.com/developers/carto-vl/reference/#cartoexpressionsanimation), defaults to [0.15, 0.15]
+`variableName` Name for the animation variable creation / lookup. Will be called `animation` by default.
+`propertyName` Name for the style property that is animated. Will be set to `filter` by default.
+
+#### AnimationControls.animation : VLAnimation
```
jesusbotella: Why don't we set a get method for that?

packages/components/src/components/as-animation-controls-widget/README.md
```diff
@@ -0,0 +1,122 @@
...
+#### **play**
+
+Triggered when the user clicks on the play button.
+
+```code
+lang: javascript
+---
+const animationControls = document.querySelector('as-animation-controls-widget');
+animationControls.addEventListener('play', event => {
+  // Start playing the animation
```
jesusbotella: Not sure about it, but for me these comments look like you should start playing the animation within the handler, rather than reacting to that event with your own behaviour.

packages/components/src/components/as-animation-controls-widget/as-animation-controls-widget.scss
```diff
@@ -2,6 +2,7 @@
 
 as-animation-controls-widget {
   --as--animation-controls-widget--background-color: var(--as--color--ui-01, #FFF);
+  --as--animation-controls__progress-value--color: var(--as--color--type-02);
```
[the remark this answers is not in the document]
elenatorro: ouch, missed that 👌

packages/bridge/src/vl/VLBridge.ts
```diff
@@ -322,7 +324,12 @@ export default class VLBridge {
       duration,
       fade,
       this._layer,
-      this._rebuildFilters
+      () => {
+        const property = propertyName ? propertyName : 'filter';
```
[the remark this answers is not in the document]
elenatorro: This callback is called only once when the AnimationControls is created, so there's no problem with having the if there or before.

```js
const {
  // ...
  propertyName = 'filter'
} = options
```

```js
() => {
  if (propertyName === 'filter') {
    this._rebuildFilters();
  }
}
```

packages/bridge/src/vl/animation-controls/AnimationControls.ts
```diff
@@ -46,6 +50,10 @@ export class AnimationControls {
     }
   }
 
+  public fn(x: any): any {
```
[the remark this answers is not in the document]
elenatorro: 🤦‍♂

packages/bridge/src/vl/animation-controls/AnimationControls.ts
```diff
@@ -65,21 +73,14 @@ export class AnimationControls {
   private _onLayerLoaded() {
     this._viz = this._layer.viz;
 
-    if (!this._viz.variables[this._variableName]) {
-      this._animation = this._createAnimation();
-
-      /* Big hack, this is done internally on VL */
-      this._animation.parent = this._viz;
-      this._animation.notify = this._viz._changed.bind(this._viz);
+    const expr = this._variableName && this._viz.variables[this._variableName]
```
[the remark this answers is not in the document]
elenatorro: Good catch 👀 , this is working now cause we're always using animations that are defined in the `viz`. However, when creating an animation by default, it's needed to set the parent and the notify event. I'm doing it now in the `_createAnimation` private method 👍

examples/bridge/animation-controls/date-format.html
elenatorro: It's actually in many examples, not sure why 🤔

examples/bridge/animation-controls/date-format.html
elenatorro: Nope! it's not needed in this example, good catch!

packages/bridge/src/vl/animation-controls/README.md
elenatorro: It's a getter now 👌

packages/bridge/src/vl/animation-controls/AnimationControls.ts
elenatorro: Actually, after taking a look, they're only being used in the Animation Controls.

packages/bridge/src/vl/VLBridge.ts
```diff
@@ -457,17 +465,28 @@ export default class VLBridge {
...
 
-    if (this._animation && this._animation.propertyName === 'filter') {
-      newFilter = `@${this._animation.variableName} and ${newFilter}`;
+    if (this._layer.viz.filter.isAnimated()) {
+      const filterExpr = this._layer.viz.filter;
+      const animationFilter = VL_BINARY_EXPRESSION_TYPES.indexOf(filterExpr.expressionName) > -1
+        ? filterExpr.a.expressionName === 'animation' ? filterExpr.a : filterExpr.b
+        : filterExpr;
+
+      // TODO change animation.toString()
```
elenatorro: Note: CARTO VL issue https://github.com/CartoDB/carto-vl/issues/1381

Jesus89: It would be nice if you can add the link of the issue in the TODO

Jesus89:
I found a little issue because the slider overflows in the animation control:



This can be easily solved by removing `overflow: auto` from the `as-animation-controls-widget` class.

Jesus89:
There is another mini-issue. When the header is empty it occupies a blank space:

Jesus89:
The `as-widget-header` adds still a blank space when there is no title/description

PR #636, opened 2019-07-10
Fix Time Series widget

elenatorro:
Fix https://github.com/CartoDB/cartoframes/issues/813
---

This PR allows to define a Time Series Widget using also a style property (like `filter`) instead of having to declare it as a viz variable.


packages/bridge/src/vl/time-series/TimeSeries.ts
```diff
@@ -174,7 +166,23 @@ export class TimeSeries {
...
 
-  private _createAnimation() {
+  private _getAnimationExpression() {
+    if (this._variableName && this._viz.variables[this._variableName]) {
+      return this._viz.variables[this._variableName];
+    }
+
+    const expr =  this._propertyName && this._viz[this._propertyName].isAnimated()
+      ? this._viz[this._propertyName]
+      : this._createDefaultAnimation();
```
Jesus89: Could we use in the `else` the following expression? `viz[prop] AND defaultAnimation`?

packages/bridge/src/vl/histogram/CategoricalHistogramFilter.ts
```diff
@@ -50,7 +51,7 @@ export class CategoricalHistogramFilter extends BaseHistogramFilter<string[]> {
    * @memberof CategoricalHistogramFilter
    */
   public get filter(): string {
-    if (this._selection === null) {
+    if (this._selection === null || this._isTimeSeries) {
```
rjimenezda: Can this happen? Categorical histograms are going to have data that looks like: 

```
{
category: <string>
value: <number>
}
```

packages/bridge/src/vl/histogram/NumericalHistogramFilter.ts
```diff
@@ -172,8 +181,8 @@ export class NumericalHistogramFilter extends BaseHistogramFilter<[number, numbe
     if (evt.detail === null) {
       this._selection = null;
     } else {
-      const selection = (this._isTimeSeries ? evt.detail : evt.detail.selection) as [number, number];
-      this._selection = [Number(selection[0]), Number(selection[1])];
+      const selection = (this._isTimeSeries ? evt.detail : evt.detail.selection) as [number , number];
```
rjimenezda: I think Tslint should complain about this space but I dunno, it might not be ran on CI

```suggestion
      const selection = (this._isTimeSeries ? evt.detail : evt.detail.selection) as [number, number];
```

packages/components/src/components/as-histogram-widget/as-histogram-widget.tsx
```diff
@@ -365,7 +365,7 @@ export class HistogramWidget {
    * @memberof HistogramWidget
    */
   @Method()
-  public async getSelection(): Promise<number[] | string[]> {
+  public async getSelection(): Promise<Array<number | Date> | string[]> {
```
rjimenezda: I know typescript likes to complain about [] vs Array, but I wonder if we could unify this to Array<number | Date | string> ?

packages/components/src/components/as-histogram-widget/interfaces.d.ts
```diff
@@ -1,6 +1,6 @@
 export interface HistogramData {
-  start: number;
-  end: number;
+  start: Date | number;
```
rjimenezda: This is nitpicky, and I wouldn't really try changing it, but this is tecnically incorrect. If start is a Date, end must be a date too.

The solution is probably making HistogramData a generic HistogramData<T>, but again, don't even bother imho.

packages/components/src/components/as-histogram-widget/as-histogram-widget.tsx
```diff
@@ -563,13 +563,15 @@ export class HistogramWidget {
 
     this.barsContainer = drawService.renderPlot(this.container);
 
+    const tooltipFormatter = this.tooltipFormatter || this._tooltipFormatter;
```
rjimenezda: I think this is never going to be `this._tooltipFormatter`, this.tooltipFormatter has a default value of this.formatter.

packages/components/src/components/as-time-series-widget/as-time-series-widget.tsx
```diff
@@ -384,7 +395,7 @@ export class TimeSeriesWidget {
       }
 
       // We have to coerce to number[] because it can also be string[] for categorical histograms
```
rjimenezda:
```suggestion
      // We have to coerce to Array<number | Date> because it can also be string[] for categorical histograms
```

packages/components/src/components/as-histogram-widget/as-histogram-widget.tsx
rjimenezda: If you think this behaviour is better (showing the start + the value) we should change it in this.formatter, but keep in mind that one does some differentiation between categorical data and numerical data.

BTW, while this is a change in the default behaviour of the component I think it doesn't qualify as a breaking change, since users are able to change it in case it's not exactly what they want.

packages/components/src/components/as-time-series-widget/as-time-series-widget.tsx
```diff
@@ -433,8 +444,12 @@ export class TimeSeriesWidget {
...
   }
 
-  private axisFormatter(value: number): string {
-    return this._formatter(new Date(value));
+  private axisFormatter(value: number | Date): string {
+    return this._formatter(value);
+  }
+
+  private _tooltipFormatter(data: TimeSeriesData): string {
+    return `${data.start}, ${data.value}`;
```
rjimenezda: You can make multi-line tooltips by returning an array of strings, just in case you think it's better to show a line for the start value and then the value itself below.

packages/components/src/components/as-time-series-widget/as-time-series-widget.tsx
```diff
@@ -267,13 +274,17 @@ export class TimeSeriesWidget {
     if (sameData(newData, oldData)) {
       return;
     } else {
-      this._data = prepareData(newData);
+      this._data = newData;
```
rjimenezda: This is great and probably improves the performance a bit :clap:

The question that remains is, do we actually need a this._data then? It's not a huge overhead (just another ref to the same data), but I think we can remove it for simplicity's sake.

packages/components/src/components/as-time-series-widget/as-time-series-widget.tsx
```diff
@@ -1,16 +1,23 @@
...
+} from 'd3-time-format';
 import { icon } from '../../utils/icons';
 import { AxisOptions, HistogramColorRange, HistogramData, HistogramSelection } from '../as-histogram-widget/interfaces';
 import { RenderOptions } from '../as-histogram-widget/types/RenderOptions';
 import {
+  AUTO_FORMAT,
   DEFAULT_BACKGROUND_BAR_COLOR,
-  DEFAULT_BAR_COLOR
+  DEFAULT_BAR_COLOR,
+  DEFAULT_DATE_FORMAT,
```
rjimenezda: You should be using this as the default value of the prop, I can't make suggestions on unchanged lines :[

packages/components/src/components/as-time-series-widget/as-time-series-widget.tsx
```diff
@@ -267,13 +274,17 @@ export class TimeSeriesWidget {
     if (sameData(newData, oldData)) {
       return;
     } else {
-      this._data = prepareData(newData);
+      this._data = newData;
+
+      this.timeFormat = this.timeFormat === AUTO_FORMAT
+        ? typeof this.data[0].start === 'number' ? DEFAULT_NUMBER_FORMAT : DEFAULT_DATE_FORMAT
```
rjimenezda: This is great too, falling back to `%Q` if it's numbers <3

packages/bridge/src/vl/time-series/TimeSeries.ts
rjimenezda: I don't understand this comment, it might be outdated by now :eyes: 

You can't use viz[this._propertyName] in the else clause cause it might be null-ish, right?

examples/bridge/time-series/simple.html
```diff
@@ -63,7 +63,7 @@
     const source = new carto.source.SQL(`select * from cordoba_catastro_simple where year > 1900 and year < 2018`);
     const viz = new carto.Viz(`
       @duration: 30
-      @animation: animation(linear($year, 1900, 2017), @duration, fade(0.1, ${Number.MAX_SAFE_INTEGER}))
+      filter: animation(linear($year, 1900, 2017), @duration, fade(0.1, ${Number.MAX_SAFE_INTEGER}))
```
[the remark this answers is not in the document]
elenatorro: It's working now both in the time series and the animation controls widget. It'd actually be great to unify the methods of these widgets since they're doing mainly the same. This example works, and the other example works as well.

packages/bridge/src/vl/time-series/TimeSeries.ts
elenatorro: you're right, it can be if the property name is not a style property 👌

packages/components/src/components/as-histogram-widget/as-histogram-widget.tsx
elenatorro: yep!

rjimenezda:
Closed in favor of #638

PR #640, opened 2019-07-18
Fix time series improvements vl

elenatorro:


packages/bridge/src/vl/time-series/TimeSeries.ts
```diff
@@ -174,31 +190,26 @@ export class TimeSeries {
...
     const s = this._carto.expressions;
-    const animation = s.animation(
-      s.linear(
-        s.prop(this._columnName),
-        s.globalMin(s.prop(this._columnName)),
-        s.globalMax(s.prop(this._columnName))
-      ),
+
+    return s.animation(
+      s.linear(s.prop(this._columnName)),
```
rjimenezda: s.linear does the globalMin and globalMax by default, right?

elenatorro: Exactly 👌

PR #651, opened 2019-09-09
Fix Airship build

elenatorro:
Fixes Airship build by fixing the Stencil version.

### DIFF
--- a/packages/bridge/tsconfig.json
+++ b/packages/bridge/tsconfig.json
@@ -23,6 +23,7 @@
   ],
   "exclude": [
     "node_modules",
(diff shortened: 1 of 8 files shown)


packages/bridge/tsconfig.json
```diff
@@ -16,13 +16,15 @@
     "jsx": "react",
     "jsxFactory": "h",
     "outDir": "dist",
-    "rootDir": "src"
+    "rootDir": "src",
+    "skipLibCheck": true
```
rjimenezda: This should not be required anymore with the fixed versions.

PR #661, opened 2020-01-08
Fix animation filter category selected

elenatorro:
Fix https://github.com/CartoDB/cartoframes/issues/1359 and https://github.com/CartoDB/cartoframes/issues/1412

This PR:
- Restarts the animation when a filter is applied
- Fixes a dependent issue due to a broken d3 dependency ([more info here](https://github.com/d3/d3-brush/issues/64))
- Fixes a typescript error when importing the `mitt` event library


elenatorro:
ok 👍 There's a lint issue I still have to solve but it's an easy one

packages/bridge/src/vl/base/BaseFilter.ts
```diff
@@ -1,4 +1,6 @@
-import mitt from 'mitt';
+import { Emitter } from 'mitt';
```
jesusbotella: I think we're importing this dependency twice :(
It would be better to use `import { Emitter }, mitt from 'mitt';` or any other statement but importing it only once

elenatorro:
**Note:** this PR is a bit dirty and I've opened a new one in order to follow conventional commits and the Airship guidelines: https://github.com/CartoDB/airship/pull/662

PR #676, opened 2020-02-12
Fix: unable to select highest value in a dataset [ch58611]

elenatorro:
In CARTOframes, it is not possible to select the highest value in a Histogram cause the Numerical Histogram Filter was not including it:

filter

This PR also includes a minor fix to check if an animation exists before it tries to restart it.

### DIFF
--- a/packages/bridge/src/vl/histogram/NumericalHistogramFilter.ts
+++ b/packages/bridge/src/vl/histogram/NumericalHistogramFilter.ts
@@ -76,7 +76,7 @@ export class NumericalHistogramFilter extends BaseHistogramFilter<Array<number |
     const min = minN instanceof Date ? `date('${minN.toISOString()}')` : minN;
     const max = maxN instanceof Date ? `date('${maxN.toISOString()}')` : maxN;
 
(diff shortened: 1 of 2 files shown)


packages/bridge/src/vl/histogram/NumericalHistogramFilter.ts
```diff
@@ -76,7 +76,7 @@ export class NumericalHistogramFilter extends BaseHistogramFilter<Array<number |
     const min = minN instanceof Date ? `date('${minN.toISOString()}')` : minN;
     const max = maxN instanceof Date ? `date('${maxN.toISOString()}')` : maxN;
 
-    return `(@${this.columnPropName} >= ${min} and @${this.columnPropName} < ${max})`;
+    return `(@${this.columnPropName} >= ${min} and @${this.columnPropName} <= ${max})`;
```
Jesus89: :boom:

elenatorro: Yep, at the end it was easier than expected, but I'm afraid this might broke other things and I'd like to take a second round look 👀

PR #687, opened 2020-04-16
Improve conditional number format for histograms [ch59951]

neokore:
Fix https://app.clubhouse.io/cartoteam/story/59951/histogram-shows-incorrect-numeric-values
---

### DIFF
--- a/packages/components/src/components/as-histogram-widget/as-histogram-widget.e2e.ts
+++ b/packages/components/src/components/as-histogram-widget/as-histogram-widget.e2e.ts
@@ -91,6 +91,30 @@ describe('as-histogram-widget', () => {
 
       expect(actual.innerText).toEqual('Radio Gaga');
     });
+
+    it('should format X axis ignoring decimals and SI prefixes when range is higher than 1', async () => {
+      const element: E2EElement = await page.find('as-histogram-widget');
+
+      element.setProperty('data', histogramDataFloatingToDecimal);
+      await page.waitForChanges();
+
+      const ticks = await page.findAll('.x-axis text');
+      const ticksValues = ticks.map((tick) => tick.innerText);
+
+      expect(ticksValues).toEqual(['-0.12', '0.88', '1.9', '2.9', '3.9']);
+    });
+
+    it('should format X axis using SI prefixes when range is lower than 1', async () => {
+      const element: E2EElement = await page.find('as-histogram-widget');
+
+      element.setProperty('data', histogramDataFloatingToSIPrefix);
+      await page.waitForChanges();
+
+      const ticks = await page.findAll('.x-axis text');
+      const ticksValues = ticks.map((tick) => tick.innerText);
+
+      expect(ticksValues).toEqual(['120m', '320m', '520m', '710m', '910m']);
+    });
   });
 });
 
@@ -99,4 +123,18 @@ const histogramData = [
   { start: 10, end: 20, value: 10 },
   { start: 20, end: 30, value: 15 },
   { start: 30, end: 40, value: 20 },
+];
+
+const histogramDataFloatingToDecimal = [
+  { start: -0.1234, end: 0.1432, value: 5 },
+  { start: 1.2345, end: 1.4543, value: 10 },
+  { start: 2.5678, end: 2.7876, value: 15 },
+  { start: 3.8901, end: 3.9081, value: 20 },
+];
+
+const histogramDataFloatingToSIPrefix = [
+  { start: 0.1234, end: 0.1432, value: 5 },
+  { start: 0.2345, end: 0.4543, value: 10 },
+  { start: 0.5678, end: 0.7876, value: 15 },
+  { start: 0.8901, end: 0.9081, value: 20 },
 ];
--- a/packages/components/src/components/as-histogram-widget/utils/draw.service.ts
+++ b/packages/components/src/components/as-histogram-widget/utils/draw.service.ts
@@ -299,13 +299,13 @@ function _delayFn(_d, i) {
 function getFloatPrecision(value) {
   const expValue = value.toString();
   const expPos = expValue.indexOf('.');
-  return expPos > -1 ? expValue.slice(expPos + 1).length : 0;
+  return expPos > -1 ? expValue.length - (expPos + 1) : 0;
 }
 
 function getDomainPrecision(domain) {
   let domainPrecision = 0;
   if (!(domain[0] instanceof Date)) {
-    const domainDiff = domain[domain.length - 1] as any - domain[0];
+    const domainDiff = domain[domain.length - 1] as number - domain[0];
     const domainDiffPrecision = getFloatPrecision(domainDiff);
     if (domainDiff > 1 && domainDiffPrecision > 1) {
       domainPrecision = 1;
--- a/packages/components/src/components/as-histogram-widget/utils/draw.service.ts
+++ b/packages/components/src/components/as-histogram-widget/utils/draw.service.ts
@@ -297,7 +297,7 @@ function _delayFn(_d, i) {
 }
 
 function getFloatPrecision(value) {
-  const expValue = value.toPrecision();
+  const expValue = value.toString();
   const expPos = expValue.indexOf('.');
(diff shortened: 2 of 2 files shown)


packages/components/src/components/as-histogram-widget/utils/draw.service.ts
```diff
@@ -293,9 +296,31 @@ function _delayFn(_d, i) {
   return i;
 }
 
-export function conditionalFormatter(value) {
-  if (value > 0 && value < 1) {
-    return decimalFormatter(value);
+function getFloatPrecision(value) {
+  const expValue = value.toPrecision()
```
VictorVelarde, line 300:
```suggestion
  const expValue = value.toString()
```

It looks like without a parameter, `.toString()` is equivalent, right? **https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toPrecision**

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
```diff
@@ -293,9 +296,31 @@ function _delayFn(_d, i) {
...
+  const expPos = expValue.indexOf('.')
+  return expPos > -1 ? expValue.slice(expPos + 1).length : 0
+}
+
+function getDomainPrecision(domain) {
```
VictorVelarde, line 305: It looks like this logic deserves a bit of testing. Why don't we create some basic specs on formatting, exemplifying several combinations of values & domains?

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
neokore: Yes, they are equivalent, it doesn't change anything so it could be replaced if you prefer

neokore:
Changed `toPrecision` to `toString` and added some tests

packages/components/src/components/as-histogram-widget/as-histogram-widget.e2e.ts
```diff
@@ -91,6 +91,30 @@ describe('as-histogram-widget', () => {
...
+    it('should format X axis ignoring decimals and SI prefixes when range is higher than 1', async () => {
+      const element: E2EElement = await page.find('as-histogram-widget');
+
+      element.setProperty('data', histogramDataFloatingToDecimal);
+      await page.waitForChanges();
+
+      const ticks = await page.findAll('.x-axis text');
+      const ticksValues = ticks.map((tick) => tick.innerText);
+
+      expect(ticksValues).toEqual(['-0.12', '0.88', '1.9', '2.9', '3.9']);
```
VictorVelarde, line 104: Nice! I was thinking more on unit testing, than e2e but it works, so 👌

packages/components/src/components/as-histogram-widget/as-histogram-widget.e2e.ts
```diff
@@ -91,6 +91,30 @@ describe('as-histogram-widget', () => {
...
+    it('should format X axis using SI prefixes when range is lower than 1', async () => {
+      const element: E2EElement = await page.find('as-histogram-widget');
+
+      element.setProperty('data', histogramDataFloatingToSIPrefix);
+      await page.waitForChanges();
+
+      const ticks = await page.findAll('.x-axis text');
+      const ticksValues = ticks.map((tick) => tick.innerText);
+
+      expect(ticksValues).toEqual(['120m', '320m', '520m', '710m', '910m']);
```
VictorVelarde, line 116: I guess it is, but has this formatting logic been proposed / validated somehow on 'the-product-side'?

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
```diff
@@ -293,9 +296,31 @@ function _delayFn(_d, i) {
...
-    return decimalFormatter(value);
+function getFloatPrecision(value) {
+  const expValue = value.toString();
+  const expPos = expValue.indexOf('.');
+  return expPos > -1 ? expValue.slice(expPos + 1).length : 0;
+}
+
+function getDomainPrecision(domain) {
+  let domainPrecision = 0;
+  if (!(domain[0] instanceof Date)) {
```
jesusbotella, line 307: Just a silly question, but this is going to be always a Date instance? Checking just in case we don't have some `"2020-04-22T16:44:39.203Z"` strings.

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
```diff
@@ -293,9 +296,31 @@ function _delayFn(_d, i) {
...
+    const domainDiff = domain[domain.length - 1] as any - domain[0];
```
jesusbotella, line 308: This `any` is really a number, right?

I guess we shouldn't have mixed domains, and if we checked just above that the first element is not a Date, this should be a number.

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
```diff
@@ -293,9 +296,31 @@ function _delayFn(_d, i) {
   return i;
 }
 
-export function conditionalFormatter(value) {
-  if (value > 0 && value < 1) {
-    return decimalFormatter(value);
+function getFloatPrecision(value) {
+  const expValue = value.toString();
+  const expPos = expValue.indexOf('.');
+  return expPos > -1 ? expValue.slice(expPos + 1).length : 0;
```
jesusbotella, line 302: Why should we slice the string? We can just get the difference from the decimal position and the original string length, right?

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
neokore: Yes, you're right 😅

packages/components/src/components/as-histogram-widget/as-histogram-widget.e2e.ts
VictorVelarde: 👌 I'm ok with it. I know formatting has been a source of many questions and requests in the past, so I just wanted to be sure that we were aware of it and agreed on it from the client-side perspective.

Jesus89:
Is this ready to merge then?

neokore:
I think so

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
Jesus89: @neokore is this already done?

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
Jesus89: @neokore could you check that?

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
neokore: Yes, using the e2e tests I showed to you and @VictorVelarde agrees (https://github.com/CartoDB/airship/pull/687#discussion_r413124631)

packages/components/src/components/as-histogram-widget/utils/draw.service.ts
neokore: Yes, domain is type Domain defined as follows: 
image
Also, this domain is used in a `scaleLinear` object, that doesn't support text values (AFAIK).

Anyway, if you prefer we can check just if it's a number using `Number.isFinite` or just checking the type with `typeof`.

jesusbotella: Yes, I saw that type definition but we cannot perform type checking in values coming from VL at runtime, so that is why I was asking this :)

neokore: Changed to check `Number.isFinite`, but according to MDN, this is not supported by IE and `tsconfig` is not set to `es2017` for components and `es5` for the bridge, so unless IE support were dropped from our side, I'll suggest setting components target to `es5` too.

VictorVelarde: We're not supporting IE11 in VL also, so I would exclude it explicitly

PR #688, opened 2020-04-17
Allow multi-selection in category widget

neokore:
Story details: https://app.clubhouse.io/cartoteam/story/58614


neokore:
Using `null` instead of `0` causes the same behavior, it shows 0 for non-selected items. Maybe we could change its behavior on Airship avoiding showing values when it comes as `null`, but it will affect all Airship users until we use a flag. Maybe I would go with nulls showed as zeroes for this release and then think what would be best behavior for `null` values. What do you think?

packages/bridge/src/vl/utils/conversion/category/index.ts
```diff
@@ -10,11 +10,16 @@ import { findColorForCategory } from '../histogram';
  * @returns {object[]}
  */
 export function vlToCategory(histogram: VLCategoricalHistogram, legendData: any = []): object[] {
-  return histogram.value.map((d) => ({
-    color: findColorForCategory(d.x, legendData),
-    name: d.x,
-    value: d.y
-  }));
+  const fullCategories = (histogram as any)._categories;
```
jesusbotella, line 13: Hmm, we're using a private property here. We need to be careful as it can change suddenly without no one noticing that is being used here. I guess that having a public property here is difficult but that's what we need to achieve.

Moreover, I think we should avoid using `any` at all times. If we're gonna use `_categories` here we should add it to the current type.

packages/bridge/src/vl/utils/conversion/category/index.ts
```diff
@@ -10,11 +10,16 @@ import { findColorForCategory } from '../histogram';
...
+  const values = histogram.value;
+  return fullCategories.map((d) => {
+    const value = values.find((elem) => elem.x === d.name);
+    return {
+      color: findColorForCategory(d.name, legendData),
+      name: d.name,
+      value: value ? value.y : 0
```
jesusbotella, line 20: I agree with @Jesus89 that 0 is not the proper value for a category that doesn't exist within the current viewport. We can make some tweaks to the category widget and avoid printing the value if it is undefined or something like that.

packages/bridge/src/vl/utils/conversion/category/index.ts
```diff
@@ -10,11 +10,16 @@ import { findColorForCategory } from '../histogram';
...
 export function vlToCategory(histogram: VLCategoricalHistogram, legendData: any = []): object[] {
-  return histogram.value.map((d) => ({
-    color: findColorForCategory(d.x, legendData),
-    name: d.x,
-    value: d.y
-  }));
+  const fullCategories = (histogram as any)._categories;
+  const values = histogram.value;
+  return fullCategories.map((d) => {
+    const value = values.find((elem) => elem.x === d.name);
```
jesusbotella, line 16: Just a performance quirk, it'd be better to set an object with category name as a key and the value as the actual object value. That way we will just iterate once over all object.

packages/bridge/src/vl/utils/conversion/category/index.ts
neokore: To get this done we would need to type everything from histogram to the top, because right now `_dataLayer`, `carto`, `layer`, the widgets,... all of them are set to `any`. I could add `_categories` to `VLCategoricalHistogram` interface, but it would be just a patch that would hide the real problem.
IMO, it would be great to do it, but it deserves its own task to focus and set the right types everywhere it's needed, and not just here; maybe blocking this feature until then if you think that's needed.

packages/bridge/src/vl/utils/conversion/category/index.ts
neokore: Sure, but will it change for everybody or will we set a flag/option to define this behavior in order to avoid breaking things to current Airship users in a minor version?

packages/bridge/src/vl/utils/conversion/category/index.ts
neokore: Totally, my fault!

packages/bridge/src/vl/utils/conversion/category/index.ts
jesusbotella: Yes, I mean, it would be great to have everything with the proper typing but we'd have needed to start since the very beginning.

What I meant is that `(histogram as any)` cast is not good, and even if `_categories` is a private property we should add it to the typing if we're going to use it.

packages/bridge/src/vl/utils/conversion/category/index.ts
jesusbotella: I don't know, my take would be to do the change for all users. It's not a major change IMHO.

If value is undefined, no value is shown.
If value is null, 0 is shown.
If value is a number, the number is shown then.

But let's see what other participants in the PR think.

Jesus89: If undefined does the trick, let's go for it. IMO

packages/bridge/src/vl/utils/conversion/category/index.ts
neokore: Ok, not I understand. I put this cast to emphasize that private variable access; but I think you are right, maybe it'd be better to set it in the interface with a comment and create a task to make it  "official" until it is broken in the future.

packages/bridge/src/vl/utils/conversion/category/index.ts
neokore: Didn't try with `undefined` yet, but anyway, if any user currently uses `undefined` and expects a 0 it will not be like this anymore, so this tricky behavior needs at least be documented, though I wouldn't recommend this `null` vs `undefined` trick unless it is more or less standard.
Anyway, I will try it.

cmongut: I would go for it too

neokore: `Undefined` doesn't work, it draws `NaN` in the widget and somehow breaks sorting items. Not bad as long as nobody will be using this and it should be ease to fix, though I still don't recommend this trick, I rather treat `undefined` and `null` values the same or add a `hideNulls` like flag.
Anyway, both options should work, so which do you prefer?

VictorVelarde: I completely agree on this. I would not use undefined and null with those different meanings

IMO:

If value is null, no value is shown.
If value is a number, the number is shown then (also if we have a true 0, of course)

If value is undefined,.... Do we ever get undefined? Why?

packages/bridge/src/vl/utils/conversion/category/index.ts
Jesus89: I agree with @jesusbotella. We should find a better way to access private categories in this case. What about adding a getter to the VLCategoricalHistogram class "getAllCategories"?

packages/bridge/src/vl/utils/conversion/category/index.ts
Jesus89: Agree. I think we tried undefined instead of null to check if it does the trick, but if it doesn't work we should use a defined null to set "empty" value and then fix the widget to support that

packages/bridge/src/vl/utils/conversion/category/index.ts
neokore: It is an interface, I rather type what's coming from this (`dataLayer.viz.variables`) from its library, and set `categories` as a public property or add a public method to get this categories.

Jesus89: @jesusbotella @VictorVelarde what do you think about this? It's OK for you (for now) to use `_categories` directly, or with a method. With that decision, we can unblock the PR and merge it

VictorVelarde: I guess we all prefer a public method, to officialize that it is accessed from outside. I don't have a strong opinion its the naming, so as you wish

VictorVelarde:
LGTM if we change the private property in the interface to something public, wether 'categories', 'getAllCategories' or your preference

PR #694, opened 2020-05-20
fix: remove SI format and set new rules to format X axis for histogram [ch76214]

neokore:
Fix [ch76214]
---

### DIFF
--- a/packages/components/src/components/as-histogram-widget/as-histogram-widget.e2e.ts
+++ b/packages/components/src/components/as-histogram-widget/as-histogram-widget.e2e.ts
@@ -104,7 +104,9 @@ describe('as-histogram-widget', () => {
       expect(ticksValues).toEqual(['-0.12', '0.88', '1.9', '2.9', '3.9']);
     });
 
-    it('should format X axis using up to 3 decimals when range is lower than 1 using values with precision lower than 3', async () => {
+    it(
+      'should format X axis using up to 3 decimals when range is lower than 1 using values with precision lower than 3',
+      async () => {
       const element: E2EElement = await page.find('as-histogram-widget');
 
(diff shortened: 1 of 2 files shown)


packages/components/src/components/as-histogram-widget/as-histogram-widget.e2e.ts
```diff
@@ -104,16 +104,57 @@ describe('as-histogram-widget', () => {
       expect(ticksValues).toEqual(['-0.12', '0.88', '1.9', '2.9', '3.9']);
     });
 
-    it('should format X axis using SI prefixes when range is lower than 1', async () => {
+    it(
+      'should format X axis using up to 3 decimals when range is lower than 1 using values with precision lower than 3',
```
VictorVelarde, line 108: it says 'using up to 3 decimals' but the assert displays just up to 2. Should we rename it? add more elements to histogram to get an example with that 3 ?

packages/components/src/components/as-histogram-widget/as-histogram-widget.e2e.ts
```diff
@@ -132,9 +173,30 @@ const histogramDataFloatingToDecimal = [
...
+];
+
+const histogramDataFloatingToDecimalPrecision6 = [
+  { start: 0.12341234, end: 0.14321432, value: 5 },
+  { start: 0.23452345, end: 0.45434543, value: 10 },
+  { start: 0.56785678, end: 0.78767876, value: 15 },
+  { start: 0.89018901, end: 0.90819081, value: 20 },
+];
+
+const histogramDataFloatingToDecimalMicro = [
```
VictorVelarde, line 197: A picky thing: if this data constants are just used in one spec, IMO they should be inside of it, near the asserts. That way it would be easier to quickly understand the spec without scrolling.

PR #707, opened 2020-08-07
feat: Add option `show-clear-button` for histogram and time series widgets

manmorjim:
Add the parameter `show-clear-button` and mark as deprecated the `show-clear` option for histogram and time series widget.
This make those widgets having the same api as the rest of widgets.


VictorVelarde (approved):
🚀

PR #708, opened 2020-08-07
feat: Add width & height parameters to Infowindow component

manmorjim:
Add the option `width` to allow users provide custom CSS value for the width of the content.
addition: + and also `height`


VictorVelarde (approved):
Nice job. I have added also height, just in case

PR #723, opened 2020-11-04
Prop types fix

alasarr:


packages/react-ui/src/widgets/CategoryWidgetUI.js
```diff
@@ -107,7 +107,7 @@ CategoryWidgetUI.propTypes = {
   data: PropTypes.arrayOf(
     PropTypes.shape({
       category: PropTypes.string.isRequired,
-      value: PropTypes.number.isRequired
+      value: PropTypes.number
```
neokore, line 110: ~Here we can use `oneOfType` and allow number/string/null~
After checking the docs again, I see that this is the recommended way for `null` (plus `undefined` and so)
I agree with the change.

packages/react-ui/src/widgets/FormulaWidgetUI.js
```diff
@@ -44,12 +44,13 @@ FormulaWidgetUI.defaultProps = {
...
     PropTypes.string,
+    PropTypes.number,
     PropTypes.shape({
       value: PropTypes.oneOfType([
         PropTypes.number,
         PropTypes.string
-      ]).isRequired,
-      unit: PropTypes.string.isRequired
+      ]),
+      unit: PropTypes.string
```
neokore, line 53: ~same here, until we want to allow empty units using this format, in that case, ok~
Maybe here I'll rather improve the way to paint the unit, to skip it in case of not existing, because right now it could leave an empty space before the value.

neokore (approved):
LGTM

PR #725, opened 2020-11-04
passing props to makestyle

Josmorsot:


packages/react-ui/src/widgets/WrapperWidgetUI.js
```diff
@@ -98,7 +98,7 @@ const IconMinimize = () => (<svg xmlns="http://www.w3.org/2000/svg" width="24" h
 
 function WrapperWidgetUI(props) {
   const wrapper = createRef();
-  const classes = useStyles();
+  const classes = useStyles(props);
```
VictorVelarde, line 101: Out of curiosity, what kind of props are we expecting?

VictorVelarde (approved):

packages/react-ui/src/widgets/WrapperWidgetUI.js
neokore: +1

Josmorsot: this is necessary to fix the issue with the `cursor`. In styles we are checking if `expandable` is **true** but it's not working because props is empty.

https://github.com/CartoDB/airship/blob/7970605552eacbd88bf59f6a6c430ec6b98eb55a/packages/react-ui/src/widgets/WrapperWidgetUI.js#L47

neokore: Great fix @Josmorsot!

neokore (approved):

alasarr (approved):

PR #728, opened 2020-11-05
HistogramWidgetUI

Josmorsot:
I've added de HistogramWidgetUI. Now I'm going to work in filtering but would be useful that you take a look to the current implementation

### DIFF
--- a/packages/react-ui/src/stories/widgets/HistogramWidgetUI.stories.js
+++ b/packages/react-ui/src/stories/widgets/HistogramWidgetUI.stories.js
@@ -9,64 +9,52 @@ export default {
 
 const DATA = [
   {
-    "revenue": 170155,
-    "date": "2019-01-01",
-    "avg": 124866.42434721242
+    "value": 170155,
+    "tick": "2019-01-01"
   },
   {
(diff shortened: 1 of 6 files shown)


packages/react-ui/src/stories/widgets/HistogramWidgetUI.stories.js
```diff
@@ -0,0 +1,93 @@
...
+import HistogramWidgetUI from '../../widgets/HistogramWidgetUI';
+
+// This default export determines where your story goes in the story list
+export default {
+  title: 'Widgets/HistogramWidgetUI',
+  component: HistogramWidgetUI,
+};
+
+const DATA = [
+  {
```
alasarr, line 11: This is quite generic for our use case.

Something like this will be better:
```javascript
[{
  "tick": "2019-01-01"
  "value": 170155
}]
```

alasarr (changes requested):
Data format

alasarr (changes requested):
https://github.com/CartoDB/cra-template-carto/pull/55

alasarr (approved):

